PR #2, opened 2017-06-21
Ginga plugin example

mkelley:
This is a partially-functional example of a Ginga plugin for cometary image enhancements.  With Ginga version 2.6.2 or later, the plugin should be automatically discovered once sbpy is installed.

To use it, start up `ginga`, open an image of a comet (for example the test image at http://www.psi.edu/research/cometimen), and select the "Operation" button just under the image.  A menu will open; choose CometaryEnhancements.  Enter the x and y center of the comet (the pick and centroid buttons do not work, they are just for show), and select the "Enhance" button.  A new channel should be opened with a 1/rho enhancement applied to the image, e.g.,:

ginga-cometary-enhancements

This took me several hours to figure out and code up, but it should be much easier to work on going forward.  There are likely other interesting ways we can use Ginga+sbpy!


mommermi:
Thanks, Mike, this looks really cool! This could definitely be useful for the PSF subtraction tools and the maybe even the observation planning. 

Do you know if you can call a ginga session from within a script, e.g., Jupyter?

mkelley:
Yes, Ginga can be used this way, but not with all the interactive widgets. 
 There is an example notebook at: https://github.com/ejeschke/ginga/blob/26c16a869e94c5284279fe2527d4600cbb102d9d/ginga/examples/jupyter-notebook/ginga_ipython_demo.ipynb

PR #4, opened 2017-06-30
Parse comet and asteroid target names.

mkelley:
The testing is strict: letter case must be correct and there cannot be any extra characters (except leading a trailing spaces are stripped before the test).

Both functions raise an exception if the target name does not match the pattern.


mommermi:
Mike, would you mind if we move `parse_comet` and `parse_asteroid` into `data.Misc`? There is already a bunch of other naming functions there, so that would be a good match.

mkelley:
I saw the `Misc` class, but I didn't see how this was related.  `parse_comet` and `parse_asteroid` operate on strings, and return strings.  Since they don't operate on `DataClass` objects, and don't return `DataClass` objects, it didn't seem like they belong with a `Misc` data class.  I think a `misc` module would make sense though.  Thoughts?

mkelley:
Ah, I see the `Misc` class is not a `DataClass` object, but either way, if we want to wrap a bunch of various nearly unrelated functions together, I feel like they should be in a module, rather than a class.

mommermi:
I agree, let's turn `Misc` into a module then. This might also be a good place to have all the lower level `utils` functions that will pop up at some point.

mkelley:
Great, will do.

mkelley:
OK, I've declassified the `Misc` class into its constituent functions.  Because they are still related to returning data from online sources, I left them in the `data` module for the moment.

I've moved everything related to target names to a `name` module:
  * altident
  * asteroid_or_comet
  * parse_comet
  * parse_asteroid
  * from_packed
  * to_packed

The only one I'm not 100% sure of is `altident`, which is an online search function.  That could be moved back to `data` depending on your philosophy.

I also looked at all the classes and docstrings in `data` and fixed them as classmethods or normal methods.  For example, `Ephem.report_to_mpc()` was a classmethod:
```python
        >>> from sbpy.data import Ephem
        >>> eph = Ephem.from_array...
        >>> report = Ephem.report_to_mpc(eph)
```
But since it operates on ephemeris data, it is better off as a normal method of the `Ephem` class:
```python
        >>> from sbpy.data import Ephem
        >>> eph = Ephem.from_array...
        >>> report = eph.report_to_mpc()
```
See also `Orbit.orbit` and `Orbit.integrate`.

mommermi:
Thanks, Mike, I like this approach very much! I would leave `altident` in the `names` module so that we have functions sorted by topic, not by method (e.g., we don't have to collect all query functions in `data`).

PR #6, opened 2017-07-05
Bib registry

mkelley:
Draft registry system using the `Bib` class that was already there.  To test:
```python
from sbpy import bib, thermal
bib.track()
neatm = thermal.NEATM()
bib.report().to_text()
# 'sbpy.thermal.NEATM:\n  method: 1998Icar..131..291H\n'
bib.reset()
bib.stop()
bib.report().to_text()
# ''
neatm = thermal.NEATM()
bib.report().to_text()
# ''
```

### DIFF
--- a/sbpy/bib/bib.py
+++ b/sbpy/bib/bib.py
@@ -1,11 +1,45 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+================================
+SBPy Biliography Tracking Module
+================================
+
+`sbpy` classes and functions can automatically report citations to the
+user through a biliography registry.  Use `track` to enable citation
+tracking, and `citations` to report the references used.
+
+Example
+-------
+>>> from sbpy import bib, data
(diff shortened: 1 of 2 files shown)


mommermi:
Looks good, but it seems to me that the class structure is not necessary anymore. I propose a simplification: [bib-registry-2](https://github.com/mommermi/sbpy/blob/bib-registry-2/sbpy/bib/bib.py)

I got rid of the class `Bib` and moved `to_text()` and `to_bibtex()` outside the class. Everything now runs on the global `OrderedDict`. The syntax is pretty much the same you suggested, except that `report()` is not necessary anymore. You can call `bib.to_text()` directly, which should make it even easier for the user. Thoughts?

mkelley:
Keeping it simple for now makes sense.  With some experience we can decide if some other structure is needed in the future.  I'll merge this now and let you merge your branch.

mommermi:
done, thanks!

PR #8, opened 2017-11-17
Sun model

mkelley:
An attempt at providing a solar spectrum.  The module requires `synphot` which is a generalized version of STScI's `pysynphot` for the spectral binning.  `synphot` looks very sophisticated and was approved to be an astropy affiliated package last week, so I think it is a good choice.

We need input spectra, so I left that to STScI's reference data set: http://www.stsci.edu/hst/observatory/crds/throughput.html .  The Kurucz 1993 atlas files (synphot4.tar.gz) have the solar spectra.  The module loads a couple spectra produced by Colina et al. 1996.  Maybe we should just include these and whatever other solar spectra in our package, if permitted?

I also have a Jupyter notebook showing the spectra, and how they are rebinned to R~25.  See that for details.  There are some subtle differences between the two.


mkelley:
Convenient link to the notebook: https://github.com/mommermi/sbpy/blob/f69f880562c77ef64dca19c87743aed6c96862fc/examples/data/sun.ipynb

mkelley:
And please suggest other solar spectra to include.

mkelley:
I've added a `solar_filt` function.

@jianyangli I'm not able to reproduce the values in Li et al. 2014.  The flux densities are within a few percent, but Vega system magnitude for F606W is 0.1 mag off.  See the updated notebook for details: https://github.com/mommermi/sbpy/blob/942eb22a0835c43294c49960cd2550d1b4357b9f/examples/data/sun.ipynb

mommermi:
Mike, do you want me to merge right now, or do you want to wait for Jian-Yang?

mkelley:
Hold off for now.  I'd like to get some more confidence on its accuracy and precision first.

Also, what should we do about the solar spectra and filter transmission?  Do we prefer to leave those out of the repo, or include them?  The former is definitely easier for the user (including me!), but the drawback is that we would need to maintain them.

mommermi:
Ok, will wait. 

I think we should include the spectra and transmission data - or even better: synphot should do that. Should we ask them what they think?

Also, `sun` is currently part of `sbpy.data`. Any thoughts on moving it to `sbpy.spectroscopy'?

jianyangli:
@mkelley Sorry for my silence.  Have you resolved the difference in your values of solar flux and Vega magnitude with F606 from what I had in Li et al. (2013)?  Could the different solar spectrum and throughput be the cause?  I used E490 for solar spectrum, and wfc3_uvis_f606w_004_syn.fits for filter throughput.

jianyangli:
Regarding including solar spectra and filter transmission, my opinion is to just include a few standard or most often used ones, but leave most out of the repo.  The reason is because there are so many, and we'll get complaints if not including all and maintaining them in a timely manner.  Also those could keep changing as calibration is improved or changed.

I think `sun` should be in `sbpy.data` because it's also relevant to other modules such as `sbpy.activity`.

mommermi:
@jianyangli I think it is a good idea to provide only a basic set of spectra. In that case, we should also provide a simple interface so people can use their own favorite spectra.

I think I see what you mean concerning the location of `sun`: `data` is the default location for external and fixed data files. However, I used this name for the `sbpy.data` module, which includes the data containers `Ephem`, `Orbit`, `Phys`, and `Names`. I am still not convinced that `sun` fits in there. If you are looking for something along the lines of the original `data` directory definition, we could revive this concept with a `static_data` directory? Or maybe `utils`? Let me know what you think.

mkelley:
Sounds like `sun` should be its own sub-module, i.e., `sbpy.sun`.  It is relevant to `spectroscopy` and `activity`, but is not strictly a data module.  It also might be used in `photometry`?

mkelley:
@jianyangli I did use the E490 spectrum, and that didn't resolve the issue.  I'll check that specific filter profile.

jianyangli:
@mommermi I see.  I did misunderstood the meaning of the current `data` module.  I agree with @mkelley that it might be worth its own module, although I suspect/hope there could be a better solution.  Remember that the `sun` module also contains some filter throughput data.

mommermi:
@mkelley @jianyangli Thanks, guys, for your feedback. I see that `sun` will be used by a wide range of different submodules. I am just concerned that we will accumulate a bunch of similarly useful functions and submodules over time and I would prefer to not have those on the top-level of `sbpy`. The top level level should be reserved for the major submodules. Hence, I would prefer if we put it in a separate submodule like `utils`, or similar. That would also be a good place for static data and other useful functions.

mkelley:
We have a similar problem with the Hale-Bopp filter calibration routines.  I thought `photometry` was the best place, but, being that the filters are comet related, it was suggested that `activity` was better home.  In my personal library, I keep the Sun and other standards in a `calibration` module.  HB filter calibration routines can also go there in an `hb` submodule.  Does this sound like a better solution?

mommermi:
Ok, this is tricky. Maybe `data` isn't such a bad place for `sun` after all. We can use it for data containers and the traditional python `data`, i.e., static files. Agreed?

`hb` might still be a better match for `activity`, or even `photometry`. @mkelley you pick!

migueldvb:
I think it is fine to keep `sun` in the `sbpy.data` sub-module. To download the spectra files we could use the helper functions in `astropy.utils.data` to download and cache files from given URLs. Perhaps we can include some of the solar spedtra FITS files in the repository but I would try to avoid it because handling changes in the files would normally make the git workflow more difficult although git has some support for binary files.

mkelley:
How about converting to text-based tables?

mommermi:
Doesn't `spectutils` have something along these lines?

mommermi:
@mkelley btw, can we merge this now and keep it in `sbpy.data`?

mkelley:
Not yet.  Jian-Yang and I have almost agreed on the accuracy, and I'd like to add a few solar spectra as text-based tables.

mkelley:
I'm rethinking the `sun` module, and looking at astropy and astropy.cosmology for guidance.  Astropy has an object named `ScienceState`: "Science state subclasses are used to manage global items that can affect science results."  With this in mind, I propose the following:

* We create a `Sun` abstract base class.  All solar spectra should be derived from this class.  This class defines the `fluxd` and `filt` methods that I already wrote above.
* Methods that rely on a solar spectrum would import `default_sun` from the `sun` module, and use its `get` method to return whichever solar spectrum is in effect.
* A user would write code like:
```python
...
from astropy.data.sun import default_sun, Kurucz1993

# Use the sbpy default solar spectrum
print(Afrho(1 * u.cm).fluxd(wave, aper, eph))

# which spectrum was it?
print(default_sun.get().description)

# try out some other spectra:
with default_sun.set(Kurucz1993):
    print(Afrho(1 * u.cm).fluxd(wave, aper, eph))

with default_sun.set('Colina1996'):
    print(Afrho(1 * u.cm).fluxd(wave, aper, eph))

```

What do you think?

mkelley:
Of course, this simple example doesn't really demonstrate the benefits over a keyword argument as I already used for the `Afrho` methods.  But seeing that this is a preferred mechanism for astropy, I think it is something to seriously consider.

mommermi:
I like this approach. It seems clean to me. We can probably adopt it in different places as well, like average asteroid reflectance spectra.

mkelley:
With respect to sbpy data files, I read the following in the astropy coding guidelines: "Packages can include data in a directory named data inside a subpackage source directory as long as it is less than about 100 kB."  Just a few solar spectra will quickly fill up that 100 kB.  I'll look into the right mechanism to retrieve the data via astropy.  There is some magic in the examples that I don't yet understand.  Any suggestions, @migueldvb ?

http://docs.astropy.org/en/latest/development/codeguide.html#data-and-configuration

mkelley:
And for a user provided solar spectrum:

```python
with default_sun.set(UserSun(filename)):
    ...
```

migueldvb:
This module approach looks very helpful.

Using the `astropy.data` tools one can download the data to a local file or provide a file-like object that can be read with the `with` statement. Since the solar spectra are available inside large tar files from that contain a large number of files, we should make the individual files we need available somewhere, perhaps from the astropy data server: https://astropy.stsci.edu/data/

mkelley:
OK, I have this refactored in a new branch.  I think it will be the right way to go, so I'll close this PR and open a new one.

PR #10, opened 2017-11-18
Add photodissociation timescales for radio species

migueldvb:
Photodissociation timescales for HCN, CH3OH and H2CO taken from Crovisier 1994.


mkelley:
@migueldvb Thanks for the updates.

PR #12, opened 2017-11-28
Activity updates

mkelley:
This adds `Afrho`, `Efrho`, and `Haser` classes, some related constants, and general support structure.  Once the solar flux density functions are sorted out, this module can be finished.

I found some results in the literature to use as benchmarks, but they do not agree with the code in all cases.  I suspect they are issues with the literature, rather than the code, but I will need to investigate further.  @migueldvb, might you have some suggestions to use as testing standards?


mommermi:
Nice, thanks, guys!

PR #13, opened 2017-11-29
Bibliography functions

migueldvb:
This implements the `to_text` and `to_bibtex` functions using the ads module to interact with the ADS.  It adds a new dependency on `ads` which is pip-installable and is the unofficial client to interact with the new [API](https://github.com/adsabs/adsabs-dev-api).


mommermi:
Thanks, Miguel!

PR #20, opened 2017-12-05
Add epsabs argument for column density integrals

migueldvb:
I think this is what is intended with the reference to the epsabs parameter in the docstring of the column density function. This module starts to look great. I checked the failing tests and could not reproduce the numbers in the publication either, so I guess we will need to update those tests or use numbers from a different source.

### DIFF
--- a/sbpy/activity/gas.py
+++ b/sbpy/activity/gas.py
@@ -348,7 +348,7 @@ class GasComa(ABC):
         """
         pass
 
-    def _integrate_column_density(self, aper):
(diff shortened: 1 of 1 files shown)


mkelley:
Ah, I had epsabs as a parameter at one point.  I thought that the model could be sped up with a lower integration error tolerance.  However, it didn't feel strike me as any faster, particularly the double integral for rectangular apertures.  Anyway, I think it is OK to keep it.

mkelley:
Thanks for catching the typos and such!

PR #22, opened 2017-12-06
Update comet name to match whitespace and quote character

migueldvb:
This works for parsing comets with space or single quotes in their name as in `6P/d'Arrest` or `122P/de Vico` that I have added to the name tests. Ideally `findall` could return a single match with several groups to make it easier to parse the components (instead of a list of tuples for each group), but Iit seems that it is not really possible for comet names.

I also removed a `math` import that is not used.

### DIFF
--- a/sbpy/data/names.py
+++ b/sbpy/data/names.py
@@ -61,8 +61,6 @@ class Names():
         'J95A01A'
         """
 
-        import math
-
         if isinstance(s, dict):
             ident = s
         else:
@@ -97,7 +95,8 @@ class Names():
             return (yr + let[0] + num + let[1])
         else:
             raise TargetNameParseError(('{} cannot be turned into a '
-                                        'packed number or designation').format(s))
+                                'packed number or designation').format(s))
+
 
     @staticmethod
     def parse_comet(s):
@@ -116,7 +115,7 @@ class Names():
         r : dict
         The dictionary contains the components identified from `s`:
         number, orbit type, designation, name, and/or fragment. If
-        none of these components are identfied, a
+        none of these components are identified, a
         `TargetNameParseError` is raised
 
         Raises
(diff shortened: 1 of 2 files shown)


sbpy/data/names.py
```diff
@@ -167,15 +166,15 @@ def parse_comet(s):
...
         import re
 
         # define comet matching pattern
         pat = ('^(([1-9][0-9]*[PDCXAI]'
                '(-[A-Z]{1,2})?)|[PDCXAI]/)' # typ/number/fragm [0,1,2]
-               '|([-]?[0-9]{3,4}[ _][A-Z]{1,2}[0-9]{1,3}(-[1-9A-Z]{0,2})?)' 
+               '|([-]?[0-9]{3,4}[ _][A-Z]{1,2}[0-9]{1,3}(-[1-9A-Z]{0,2})?)'
                # designation [3,4]
-               '|(([A-Z][a-z]?[A-Z]*[a-z]*[ -]?[A-Z]?[1-9]*[a-z]*)' 
+               '|(([dA-Z][a-z\']? ?[A-Z]*[a-z]*[ -]?[A-Z]?[1-9]*[a-z]*)'
```
mkelley: Also need to support a leading lower-case 'v' for  271P/van Houten-Lemmon (according to Wikipedia).

@migueldvb , @mommermi But this makes me think, do we really want to enforce Proper Capitalization?

migueldvb: Thanks, I added a lower case v in the character set used to match the first letter of the comet name. I'm not sure if we need to ensure the right capitalization of the name. This pattern will allow several different choices of case for some names.

mkelley:
Hi Miguel, can you re-base so that we can finish this off?

migueldvb:
Excellent! I have rebased this branch on master

mommermi:
Done, thanks!

PR #23, opened 2017-12-10
Spectral and photometric calibrations

mkelley:
The PR is a re-write of #8 and allows for spectral and photometric calibrations based on solar and Vega spectra.  Two spectra are distributed with `sbpy`, the E490-00a (2014) solar spectrum standard, and a lower-resolution version thereof.  Three additional spectra are built-in: Kurucz 1993 solar spectrum, Castelli 1996 solar spectrum (I had previously called this Colina et al. 1996), and Bohlin 2014 Vega spectrum.

The `Sun` and `Vega` classes provide the spectra for the module.  They are subclasses of `SpectralStandard`.  I am not sure if this class implementation is consistent with the goals of the `Spectrum` class, so I did not connect the two.  This decision can be re-visited in the future when `Spectrum` gets implemented.

The built-in spectra are described in `spectroscopy.sun.sources` and `vega.sources`.  They can be retrieved with the `from_builtin` methods.
```python
from sbpy.spectroscopy.sun import Sun
from sbpy.spectroscopy.vega import Vega
sun = Sun.from_builtin('E490_2014')
vega = Vega.from_builtin('Bohlin2014')
```

The spectral flux densities can be inspected or rebinned:
```python
print(sun.fluxd)  # native resolution
wave = np.linspace(0.3, 0.8) * u.um
print(sun(wave))  # bins centered at wave
```
@jianyangli and I compared notes and we think the binning is working correctly now.

The default spectrum can be retrieved and controlled via `default_sun`:
```python
from sbpy.spectroscopy.sun import default_sun
with default_sun.set('Kurucz1993'):
  # Kurucz 1993 in effect

# back to module default (E490_2014)
sun = default_sun.get()
```
In a similar manner, `default_vega` controls the Vega spectrum.


mkelley:
Howdy y'all,

Is this approach OK?  Good to merge?

mommermi:
Yes, sorry! I kind of lost track of this. Will get back to sbpy once the early April proposal massacre is over...

mkelley:
Agreed!

PR #28, opened 2018-05-03
New Tracking context manager.

mkelley:
Context manager for bibliography tracking so that one can pop in and out of tracking via:
```python
from sbpy import bib, data

with bib.Tracking():
    eph = data.Ephem.from_horizons('encke', epoch=None, observatory='500')

bib.to_text()
```


mommermi:
Fancy, thanks!

PR #32, opened 2018-05-29
Add scipy and synphot dependencies to fix failed Travis CI build

migueldvb:
This should fix the Travis failure


mommermi:
Excellent, I tried looking into this but then got distracted... Thanks!

PR #33, opened 2018-05-29
 Could not parse .travis.yml 

migueldvb:
I suspect that the extra whitespace is causing the Travis build to fail again. Let's see if the build starts when the pull request is opened.


mommermi:
... looks like the merge killed some of the Travis builds. Sorry.

migueldvb:
No problem. It's running again and giving an IndentationError.

PR #34, opened 2018-05-29
Fixing Travis and docs build

migueldvb:
Hopefully this should be all that is needed.


mommermi:
@migueldvb: shall we deactivate doctests for now to simplify things?

PR #38, opened 2018-06-21
Integrate volume density along line of sight in GasComa class 

migueldvb:
This tests that the formulae in Newburn and Johnson 1978 for Haser model gives the same result as numerical integration (`test_column_density` function in `test_gas.py`).


mkelley:
That's a good idea, thanks.

Issue #31, opened 2018-05-29
Replace callhorizons calls with astroquery.jplhorizons


migueldvb:
Since we will be using astroquery for other things as well it makes sense to have it as a required dependency.

PR #39, opened 2018-06-27
Default value for observatory code in from_horizons

migueldvb:
Adding default observatory and removing unneeded center argument in `Ephem.from_horizons`.
The default observatory code from Minor Planet Center list is 500 (geocentric). This method is needed for @gianninapr's spectral fitting functions.


mommermi:
@migueldvb, let me work on your pull request to fix #31 at the same time.

migueldvb:
Thanks!

PR #51, opened 2018-08-01
updated required packages and setup script

mommermi:
updated the list of required packages for sphinx (`requirements.txt`), Travis CI (`.travis.yml`), and the setup script (located in `setup.cfg`)


mommermi:
... just realized that in #31, I require `astroquery >= 0.3.8`. Not sure if we can require specific version numbers in `setup.cfg`. I think this is only possible if we hard-code the requirements into `setup.py`. Is there any reason speaking against this?

mommermi:
Sounds like an excellent solution, especially since we will be messing a lot with astroquery over the next weeks and months. 
Can you show me how to plug in the development version instead of the pypi release?

migueldvb:
Replacing `astroquery` by `https://github.com/astropy/astroquery/archive/master.zip` in `PIP_DEPENDENCIES` should work :crossed_fingers:

mommermi:
not sure what this is (https://travis-ci.org/NASA-Planetary-Science/sbpy/jobs/410992149),  but I guess we can ignore it for now; if the rest builds, I'll merge this.

mommermi:
got it. I'm not it.

mommermi:
done. coveralls.io didn't get the build information from Travis CI...

PR #58, opened 2018-08-08
fix name resolving issues in Names/add alternative field names in DataClass

mommermi:
This pr will be used to fix names resolving issues in `sbpy.data.Names`.


mkelley:
I think that `Names.parse_comet` may be too generous.  Comet designations should always carry P/, C/, X/, or D/, or be a periodic number with a trailing P.  Rather than successfully parsing '1998 QS55', I would prefer raising a `TargetNameParseError`.

In a different manner, the MPC adopts this behavior.  Request target 'Encke' from the Ephemeris Service and it computes an ephemeris for the asteroid.  Use 'P/Encke' for the comet.  Similarly, 'PanSTARRS' returns an error because there is no asteroid named PanSTARRS https://minorplanetcenter.net/iau/MPEph/MPEph.html (but there are many comet PanSTARRSs).  Horizons recognizes the name 'Encke' is ambiguous and returns both the comet and asteroid.

What do you think?

mommermi:
@mkelley Agreed and done!

mommermi:
By accident, this PR also introduces the infrastructure for assigning alternative field names in `DataClass`. The translation lookup table (`namealts`: name -> alternatives) is located in `sbpy/data/__init__.py`.

PR #59, opened 2018-08-28
Afρ from magnitude

mkelley:
To do:
- [x] Allow for arbitrary solar apparent magnitude from the user.
- [x] Convert from Afρ to magnitude.
- [x] Report to the user the solar magnitude used (via a verbose flag?).
- [x] Integrate changes into `Efrho` class.
- [ ] Check coverage.
- [x] Document new functionality.


mkelley:
I'll come back to coverage after I modernize a few things.

PR #61, opened 2018-09-09
Photometry rebase

jianyangli:
Implementation of IAU HG models.


sbpy/photometry/core.py
```diff
@@ -5,13 +5,255 @@
...
            'PhaseFunctionModel', 'ROLOPhase',
            'ResolvedPhotometricModelClass', 'ROLO']
 
+import numpy as np
+from scipy.integrate import quad
+from astropy.modeling import FittableModel, Fittable1DModel, Fittable2DModel, Parameter
+from astropy import units
+
+
+def ref2mag(ref, radius, M_sun=None):
```
mommermi: I personally don't care, but many people prefer not to have free floating functions that are not part of a class. Should we put these functions into a general class, e.g. `Photometry`? That's what astropy does. @migueldvb, any comments on this?

sbpy/photometry/tests/test_core.py
```diff
@@ -0,0 +1,233 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+import numpy as np
+from astropy import units as u
+from astropy.modeling import Parameter
+from ..core import *
+
+def test_ref2mag():
+    assert np.isclose(ref2mag(0.1/np.pi, 460), 3.3208379018205285)
+    assert u.isclose(ref2mag(0.1/np.pi, 460*u.km), 3.3208379018205285*u.mag)
```
mommermi: `u.isclose` has been implemented in astropy version 3.0.2, causing a build fail for older versions. I suggest that you test the version of astropy and ignore the `u.isclose` test if the version is lower than 3.0.2. You can use the following code for this 

```
from distutils.version import LooseVersion
if (LooseVersion(astropy.__version__) >= LooseVersion('3.0.2')):
   ...
```

sbpy/photometry/core.py
```diff
@@ -54,90 +296,374 @@ def mag(self, phase):
...
+        TBD
 
         not yet implemented
 
         """
 
-    def distance_module(self, eph):
-        """Account magnitudes for distance module (distance from observer,
-        distance to the Sun); return modified magnitudes
+    def mag(self, pha, **kwargs):
```
mommermi: @mkelley @migueldvb @jianyangli Guys, this is a programmatic question: this specific function requires the solar phase angle as input parameter, which can be passes as a single float, a list, or an array. My question is whether we want to keep such input parameters separate, or instead just pass an `Ephem` object that holds all these properties.

If we keep the input parameters separate, it's somewhat easy to use, unless there is a large number of them (e.g., thermal modeling routines require phase angle, r, Delta...). But even in that case, we can still pull these properties from an `Ephem` object: `pha = ephem['alpha'], ...` 

If we go with `Ephem` objects instead, people will have to generate these objects, which is easy but might be perceived as unnecessary extra work if there is only one parameter required.

What is your take on this? Go with individual parameters or DataClass objects?

sbpy/photometry/core.py
mkelley: I don't understand that perspective, because they are in a sub-module, so still in some kind of namespace.  I don't see the benefit to putting these in a class.

mkelley: In other words, I am interested in hearing why they should be in classes :)

sbpy/photometry/core.py
```diff
@@ -5,13 +5,255 @@
...
+    >>> from astropy import units as u
+    >>> ref = mag2ref(2.08, 460)
+    >>> print('{0:.4}'.format(ref))
+    0.09981
+    >>> ref = mag2ref(2.08, 460*u.km)
+    >>> print('{0:.4}'.format(ref))
+    0.09981 1 / sr
+    """
+
+    if M_sun is None:
```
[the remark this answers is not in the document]
mkelley: I think that is better.  Fewer lines of code to test.

sbpy/photometry/core.py
mommermi: ... me too :)

sbpy/photometry/core.py
mkelley: Tight integration with `Ephem` is preferred.  What I've done for the `Afrho` class is accept a dictionary-like object with 'rh', 'delta', and 'phase' keys (as needed), and assumed that `Ephem` will be compatible with this.  If only one parameter is needed, then we can accept one or the other:
```python
def mag(self, pha):
  if isinstance(pha, Ephem):
    phase = pha['phase']
  else:
    phase = pha
```
Or something similar.

mommermi: Ah, ok. It's trivial if there is only parameter required: either a quantity/list/array for that parameter or `Ephem`. I was wondering about what to do in case there are several required parameters.

So we simply assume that parameters are `DataClass` objects. If you also allow for a dictionary, you can easily turn that into a `DataClass` object with `DataClass.from_dict` inside your function.

mkelley: Well, unless there is some benefit to using `DataClass`, I think allowing any dict-like object is better for users.  `Afrho.from_fluxd(wave, flam, aper, eph={'rh': 1 * u.au, 'delta': 1 * u.au, 'phase': 60 * u.deg})` is more straightforward and less command-line typing than `Afrho.from_fluxd(wave, flam, aper, eph=Ephem.from_dict({'rh': 1 * u.au, 'delta': 1 * u.au, 'phase': 60 * u.deg}))`, not to mention the additional import statement.  Here is how I documented this parameter:
```
        eph : dictionary-like or `~Ephem`
          Ephemerides of the comet, describing heliocentric and
          geocentric distances as `~astropy.units.Quantity` via
          keywords `rh` and `delta`.  Phase angle, `phase`, is
          required if `phasecor` is enabled.
```

mommermi: I agree. What I meant was that you can do the conversion from dict to `DataClass` internally in your function, saving you code lines to distinguish between a dict and `DataClass`. Just test if the input is a dict, and turn it into a `DataClass` and then assume in the remainder of your function that you are dealing with a `DataClass` object.

mkelley: Ah, of course.

sbpy/photometry/core.py
```diff
@@ -5,13 +5,255 @@
...
+        Q = True
+
+    ref = 10**((M_sun-mag)*0.4)/(np.pi*radius*radius*units.km.to('au')**2)
+    if Q:
+        return ref/units.sr
+    else:
+        return ref
+
+
+class spline(object):
```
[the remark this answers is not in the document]
jianyangli: Not really.  `scipy.interpolate.spline` doesn't work for this case, because in the G1G2 model the spline interpolation is defined with also the first derivatives at both ends, which is not in any spline implementation in Python that I could find.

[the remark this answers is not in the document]
jianyangli: Actually no.  I was not aware of `HG1G2tools` when I develop this class.  So what I did is to expand the `scipy.interpolate.spline` to include the first derivatives in both end as required by the G1G2 model.  But maybe we should instead use that `HG1G2tool` here for its computational efficiency?

sbpy/photometry/core.py
```diff
@@ -5,13 +5,255 @@
...
+            out[idx] = self.polys[-1](x[idx])
+        return out
+
+
+class DiskIntegratedModelClass(Fittable1DModel):
+    """Base class for disk-integrated phase function model
+
+    Examples
+    --------
+    - Define a linear phase function with phase slope 0.04 mag/deg, and
```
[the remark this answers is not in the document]
jianyangli: Sure, will do.
[ ] Implement a linear phase function model

sbpy/photometry/core.py
jianyangli: I thought about including it into the `DiskIntegratedModelClass`, but then considered that they should be general-purpose tools and should not belong to any class but rather as functional tools in the package.  Looks like we've come to an agreement at this point.

sbpy/photometry/core.py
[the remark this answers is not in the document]
mommermi: Ha, nice. I don't think `HG1G2tools` is any better computationally than your code, so let's go with yours. They are taking a very similar approach by defining their own splines, that's why it reminded me of it.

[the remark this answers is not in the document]
mommermi: ok, sounds good!

sbpy/photometry/core.py
```diff
@@ -5,13 +5,255 @@
...
+from astropy.modeling import FittableModel, Fittable1DModel, Fittable2DModel, Parameter
+from astropy import units
+
+
+def ref2mag(ref, radius, M_sun=None):
+    """Convert average bidirectional reflectance to reduced magnitude
+
+    Parameters
+    ----------
+    ref : num, astropy.units.Quantity
```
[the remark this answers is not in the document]
jianyangli: - [ ] replace `num`, `numbers` etc. to `float`

sbpy/photometry/core.py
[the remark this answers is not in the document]
jianyangli: The reason is because I don't want to explicitly set the value of `M_sun` in `DiskIntegratedModelClass.__init__()` or any other code that calls `ref2mag` or `mag2ref`, in order to avoid typos in the number and updating too much code if in the future the value of `M_sun` needs to be updated.

sbpy/photometry/core.py
```diff
@@ -5,13 +5,255 @@
...
+    M_sun : number, optional
+        The magnitude of the Sun, default is -26.74
+
+    Returns
+    -------
+    The same type as input, reduced magnitude (at r=delta=1 au)
+
+    Examples
+    --------
+    >>> from astropy import units as u
```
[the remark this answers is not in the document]
jianyangli: I'm not sure what you mean here.  Could you explain?  Thanks.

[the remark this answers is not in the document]
mommermi: The examples that you currently provide for all the functions do not `import` (sorry, I meant `import` not `include`; been doing some C++ recently...) your photometry module. It would be good to include this so that people know how to access the functions, just so that the examples are complete.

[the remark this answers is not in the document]
jianyangli: Got it.  Was confused by `include`.  Will do.

sbpy/photometry/core.py
jianyangli: I like the flexibility by allowing this parameter to be either array-like to pass phase angle only, or `dict` or `DataClass` to include distances.  The question is, should we do distance correction in an implicit way or an explicit way?

Personally I like an implicit way.  I.e., we perform distance correction by default.  If one distance (either solar or observer) is not provided explicitly (by either the `dict` or `DataClass` type parameter), then we simply assume it as 1 au.  What do you guys think?

jianyangli: Also, I haven't had time to follow all the previous discussions about `DataClass` and study the documents, but will just assume that `DataClass.from_dict` method will take care of the conversion of `dict` to `DataClass`, include error handling.  In my code, I will simply pass the parameter to it without any extra check.  Does this sound fair?  Thanks.

mommermi: Yes and yes. I prefer the implicit distance correction - as you said, one can override it by using distances of 1 au. Also, yes, `DataClass.from_dict` works out of the box and should be very robust. It does some error handling, if you want more features we can talk about that.

I also agree that we should be somewhat flexible in the data input, but I would encourage people to use the `DataClass` objects. If they prefer dictionaries or lists, we can use `DataClass` objects internally to keep things simple. Just yesterday I started setting up a mechanism that enables alternative property names so that, e.g., `eph['r']`, eph['r_hel`'], and `eph['heldist']` provide the same data. This means that we should think about what "official" property name we want and which ones we allow as alternative names. A while ago I setup this wiki page: https://github.com/NASA-Planetary-Science/sbpy/wiki/Data-container-content. I will populate it with what I got so far tomorrow and then we should go from there.

mkelley: I think using 1 au / 1 au as a default for the class is OK.

jianyangli:
I've updated this branch with the suggested changes.  Please check and let me know if further updates are needed.  Thanks.

mommermi:
This looks great, thanks, Jian-Yang! Could you also please create write up for the documentation similar to this: https://github.com/NASA-Planetary-Science/sbpy/blob/master/docs/sbpy/activity.rst

But maybe more importantly right now: I'll give an sbpy workshop at EPSC next week. Would you mind putting together a jupyter notebook with some short examples? No need to make it fancy or specific. Just some very simple examples that people might be interested. Have a look at this for some examples: https://github.com/NASA-Planetary-Science/sbpy-tutorial/tree/master/notebooks

jianyangli:
Ok, I’ll create a notebook with examples by Monday.

PR #68, opened 2018-09-14
support for pyoorb

mommermi:
implemented orbit transformations, orbit propagation, and ephemeris computations

docs and limited tests implemented


mommermi:
will merge this now, but added a note to the data project to add more tests later

PR #73, opened 2018-09-30
New natrual_sort_key.

mkelley:
Sorting comet names in Python does not return a pleasing order:
```python
>>> comets = ['9P/Tempel 1',
...           '101P/Chernykh',
...           '10P/Tempel 2',
...           '2P/Encke']
>>> sorted(comets)
['101P/Chernykh', '10P/Tempel 2', '2P/Encke', '9P/Tempel 1']
```
Why is 101P before 2P?  Because '101' < '2'.  Instead, the so-called natural order would sort items using numerical comparisons.  Python does not have a natural sort option, but I think we should add one to sbpy because humans like sensible comet and asteroid numbering:
```python
>>> from sbpy.data.names import natural_sort_key
>>> sorted(comets, key=natural_sort_key)
['2P/Encke', '9P/Tempel 1', '10P/Tempel 2', '101P/Chernykh']
```
How about this?

### DIFF
--- a/sbpy/data/names.py
+++ b/sbpy/data/names.py
@@ -50,7 +50,7 @@ def natural_sort_key(s):
     """
     import re
     keys = tuple()
-    for k in re.split('([0-9]+)', s):
+    for k in re.split('([0-9]+)', str(s)):
         keys += (int(k) if k.isdigit() else k,)
     return keys
 
--- a/sbpy/data/names.py
+++ b/sbpy/data/names.py
@@ -13,7 +13,46 @@ created on August 28, 2017
 from .core import DataClass
 from numpy import ndarray
 
-__all__ = ['Names', 'TargetNameParseError']
+__all__ = ['Names', 'TargetNameParseError', 'natural_sort_key']
+
+
+def natural_sort_key(s):
+    """List sort keys considering strings of numbers as integers.
+
+    Intended to be used with `list.sort` or the `sorted` built-in
+    function.
+
+
+    Parameters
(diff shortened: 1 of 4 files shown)


sbpy/data/names.py
```diff
@@ -13,7 +13,46 @@
...
+    ...           '2P/Encke']
+    >>> sorted(comets)
+    ['101P/Chernykh', '10P/Tempel 2', '2P/Encke', '9P/Tempel 1']
+    >>> sorted(comets, key=natural_sort_key)
+    ['2P/Encke', '9P/Tempel 1', '10P/Tempel 2', '101P/Chernykh']
+
+    """
+    import re
+    keys = tuple()
+    for k in re.split('([0-9]+)', s):
```
mommermi: can you replace `s` with `str(s)`, so that the sorting also works on number provided as integers?

mommermi:
This is a pretty cool and useful addition. Could you also please add some text to `docs/sbpy/data.rst`? Thanks!

mommermi:
@mkelley is this ready to be merged? I can have a look at the docs, if you want me to.

mkelley:
I'll get back to it soon.

sbpy/data/names.py
mkelley: Got it.

mkelley:
OK, if the docs look good, it is ready for merging.

mommermi:
merging this despite build errors , fix comes with #77

PR #74, opened 2018-10-03
MPC ephemerides

mkelley:
Enables `Ephem.from_mpc()`.

### DIFF
--- a/sbpy/data/ephem.py
+++ b/sbpy/data/ephem.py
@@ -219,47 +219,44 @@ class Ephem(DataClass):
         """
 
         # parameter check
-        if start is None:
-            if epochs is None:
-                epochs = Time.now()
-
-            if not np.iterable(epochs):
-                epochs = [epochs]
-        elif stop is not None:
-            if step is None:
-                raise ValueError(
-                    'step is required when start and stop are provided')
-
-            # start and stop both defined, estimate number of steps
+        if step is None:
+            _step = None
+        else:
             _step = u.Quantity(step)
             if _step.unit not in (u.d, u.h, u.m, u.s):
                 raise ValueError(
                     'step must have units of days, hours, minutes, or seconds')
+
+        if start is None:
+            if epochs is None:
+                epochs = Time.now()
+
+            if not np.iterable(epochs) or isinstance(epochs, str):
+                epochs = [epochs]
+        elif stop is not None:
+            if _step is None:
+                raise ValueError(
+                    'step is required when start and stop are provided')
+
+            # start and stop both defined, estimate number of steps
             dt = (Time(stop).jd - Time(start).jd) * u.d
-            number = int((dt / _step).decompose())
+            number = int((dt / _step).decompose()) + 1
 
         # get ephemeris
         if start is None:
             eph = []
             for i in range(len(epochs)):
+                start = Time(epochs[i], scale='utc')
                 e = MPC.get_ephemeris(targetid, location=location,
-                                      start=epochs[i], number=1, **kwargs)
-                e['Date'] = e['Date'].iso
+                                      start=start, number=1, **kwargs)
+                e['Date'] = e['Date'].iso  # for vstack to work
                 eph.append(e)
             eph = vstack(eph)
+            eph['Date'] = Time(eph['Date'], scale='utc')
         else:
-            eph = None
-            current = 0
-            eph = []
-            while number - current > 0:
-                n = min(1441, number - current)
-                e = MPC.get_ephemeris(targetid, location=location,
-                                      start=start, step=step, number=n,
-                                      **kwargs)
-                e['Date'] = e['Date'].iso
-                eph.append(e)
-                current += len(e)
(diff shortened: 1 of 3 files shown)


sbpy/data/ephem.py
```diff
@@ -127,35 +129,137 @@ def from_horizons(cls, targetids, id_type='smallbody',
         return cls.from_table(all_eph)
 
     @classmethod
-    def from_mpc(cls, targetid, epoch, observatory='500'):
-        """
-        Load ephemerides from the
+    def from_mpc(cls, targetid, epochs=None, start=None, step=None,
```
mommermi: how do you feel about allowing a dictionary with keywords `start`, `stop`, and `step` for `epochs` instead of providing these parameters on the top level? That would be the same approach I've been taking in `from_horizons` and I would like to keep the interfaces as similar as possible

sbpy/data/ephem.py
```diff
@@ -127,35 +129,137 @@ def from_horizons(cls, targetids, id_type='smallbody',
...
+            `~astropy.units.Quantity`.
+
+        stop : string or `~astropy.units.Time`, optional
+            Specifies the ephemeris end time when ``start`` is
+            provided.  Supercedes ``number``, requires ``step``.
+            Strings are parsed by `~astropy.time.Time` assuming a UTC
+            scale.  The stop epoch may not be included depending on
+            the step size and floating point arithmetic.
+
+        number : int, optional
```
mommermi: if this parameter is deprecated, shouldn't we get rid of it?

sbpy/data/ephem.py
```diff
@@ -127,35 +129,137 @@ def from_horizons(cls, targetids, id_type='smallbody',
...
         Parameters
         ----------
-        targetid : str, mandatory
-            target identifier
-        epochs : astropy Time instance or iterable, optional, default ``None``
-            epoch of elements; if ``None`` is provided, current date is used
-        observatory : str, optional, default ``'500'`` (geocentric)
-            location of observer
+        targetid : str
+            Target identifier.  Must be resolvable by the Minor Planet
```
mommermi: Could you provide a bit more information of what is resolvable by MPC?

sbpy/data/ephem.py
```diff
@@ -127,35 +129,137 @@ def from_horizons(cls, targetids, id_type='smallbody',
...
+            provided.  Supercedes ``number``, requires ``step``.
+            Strings are parsed by `~astropy.time.Time` assuming a UTC
+            scale.  The stop epoch may not be included depending on
+            the step size and floating point arithmetic.
+
+        number : int, optional
+            The number of ephemeris dates to compute.  Superceded by
+            ``stop``.
+
+        location : str, array-like, or `~astropy.coordinates.EarthLocation`, optional
```
mommermi: this is cool, I didn't know about `astropy.coordinates.EarthLocation`; will implement this too for `jplhorizons`, where currently a dictionary of coordinates is allowed

sbpy/data/ephem.py
mkelley: Done.

sbpy/data/ephem.py
mkelley: What I mean is that the input is flexible, but there is an order of precedence.  Providing ``stop`` and ``number`` at the same time over constrains the time steps, and the two parameters could conflict.  So, if ``stop`` is given, ``number`` is ignored.  As an alternative, we could raise an error if both are provided.  I definitely want ``number`` to be available for the user because this is required by `astroquery.mpc.MPC.get_ephemeris`.

sbpy/data/ephem.py
mkelley: After this is merged, we should work on getting all the overlaps into a shared code base, which will make extending `Ephem` a bit easier.

mommermi:
The crashing builds are my fault. Should be fixed now. Just restarted the Travis-CI builds for #73 and #74. Thanks!

mommermi:
D'oh! I'm on the tests...

sbpy/data/ephem.py
```diff
@@ -127,35 +129,151 @@ def from_horizons(cls, targetids, id_type='smallbody',
...
+                                      start=start, number=1, **kwargs)
+                e['Date'] = e['Date'].iso  # for vstack to work
+                eph.append(e)
+            eph = vstack(eph)
+            eph['Date'] = Time(eph['Date'], scale='utc')
+        else:
+            eph = MPC.get_ephemeris(targetid, location=location,
+                                    start=start, step=step, number=number,
+                                    **kwargs)
+
```
mommermi: one more thing: could you add a column to your resulting table named `timescale` that describes the time scale of the epoch in your table (e.g., `UTC` for UT). This is necessary for compatibility. Thanks!

mkelley: Sure.  Sounds like a check to add to Ephem.__init__?

mommermi:
merging despite build errors, fix comes with #77

Issue #66, opened 2018-09-14
Bibliography does not allow multiple references for a single key.


mkelley:
Consider that there are various estimates for the fluorescence band g-factors of CO2, e.g., Crovisier & Encrenaz 1983 vs. Debout et al. 2016.  If we included both in `activity.fluorescence_band_strength` and a user called both to determine the difference, only the last used citation will be tracked.

```python
>>> from sbpy import bib
>>> with bib.Tracking():
>>>     bib.register('task', {'g': 'reference 1'})
>>>     bib.register('task', {'g': 'reference 2'})
>>> print(bib.to_text())
task:
  g: reference 2
```

Perhaps the citations should be internally stored as lists, and new citations appended as needed.

mommermi:
fixed with #75

PR #75, opened 2018-10-05
Fix for Issue #66

gianninapr:
Fixed issue to allow the bibliography module to support more than one reference per subtask.


sbpy/bib/core.py
```diff
@@ -29,8 +29,26 @@ def register(task, citations):
     """
     global _bibliography, _track
     if _track:
-        _bibliography[task] = _bibliography.get(task, {})
-        _bibliography[task].update(citations)
+        if task in _bibliography:
+            key = []
+            value = set()
+            key.append(list(citations.keys()))
+            for i in range(len(key)):
```
mommermi: instead of working with indices, you can get the keys and values directly using something like:
`for newsubtask, newcitation in citation.items():`

sbpy/bib/core.py
```diff
@@ -29,8 +29,26 @@ def register(task, citations):
...
+                if list(citations.keys())[i] in _bibliography[task]:
```
mommermi: Instead of working with lists that are then converted to sets, you can take advantage of sets functionality immediately (e.g., adding the same element to the list will not change it). You just have to distinguish between the case that the subtask already exists or not. Can you fill in the `...` parts? Only a single line is required in each case:

> if newsubtask in _bibliography[task].keys():
> ...
> else:
> ....

sbpy/bib/core.py
```diff
@@ -29,8 +29,26 @@ def register(task, citations):
...
+                    else:
+                        value.add(list(subtasks.values())[i])
+                        value.add(list(citations.values())[i])
+                        subtasks[list(citations.keys())[i]] = value
+                else:
+                    subtasks = _bibliography[task]
+                    value.add(list(citations.values())[i])
+                    subtasks[list(citations.keys())[i]] = value
+        else:
+            _bibliography[task] = citations
```
mommermi: This part is a bit tricky: the value in `citation` is simply a string, but it should be a set. I couldn't find a better way to turn it into a set than this one:
`_bibliography[task] = {list(citation.keys())[0]: set(list(citation.values())[0])}`
Maybe you can find a better solution?

sbpy/bib/core.py
```diff
@@ -29,8 +29,26 @@ def register(task, citations):
     """
     global _bibliography, _track
     if _track:
-        _bibliography[task] = _bibliography.get(task, {})
-        _bibliography[task].update(citations)
+        if task in _bibliography:
+            key = []
```
mommermi: I think you don't have to define `key` and `value` if you are using the loop I suggest below.

sbpy/bib/core.py
```diff
@@ -29,8 +29,26 @@ def register(task, citations):
     """
```
mommermi: I think we should assume that only a single citation is provided everytime this function is called (this is the way it is setup right now) The user will not use this function manually, so we just have to call `bib.register` for each task, subtask, and reference once in our code. This will probably improve code readability. Can you change `citations` to `citation` and adjust the docstring accordingly?

sbpy/bib/core.py
```diff
@@ -89,33 +107,39 @@ def to_text():
...
-                else:
-                    # single author
-                    author = paper.first_author.split(',')[0]
-
-                output += '  {:s}: {:s} {:s}, {:s}\n'.format(key, author,
-                                                             paper.year, val)
+                if isinstance(val, str):
+                    val = [val]
+                val = list(val)
+                for i in range(len(val)):
```
mommermi: can you change this from a loop over indices to a loop over the values? Just like in my example in `bib.register`.

mkelley:
Tagging #66 .

mkelley:
Great, thanks for fixing!

mommermi:
merging despite build errors, fix comes with #77

PR #87, opened 2018-11-04
Add rh and dRA cos(Dec)

mkelley:
'rh' is my fave for heliocentric distance

dRA from JPL includes the cos(dec) term so 'dRA cos(Dec)' truly is an equivalent column name.  However, astroquery.mpc can return proper motion with and without the cos(dec) term.  This introduces an ambiguity in Ephem, and requires that dRA with and without the cos(dec) term never simultaneously exist in an ephemeris.

Is this ambiguity acceptable or do we require 'cos(dec)' when the cos(dec) term is included?  I actually think the latter is better, in which case I'll need to revise this PR.


mkelley:
@mommermi Also, velocityPA returned from Horizons is really the negative velocity vector.  Should I rename it?

mommermi:
... and here starts the mess... 

I would prefer to minimize ambiguity whenever possible. If mpc provides `dRA` and `dRA cos(dec)`, we should rename the current jplhorizons output to `dRA cos(dec)`.  It's not nice, but it might be necessary. 

Concerning `velocityPA`: yes, feel free to rename it.

One more quick note: in `data/__init__.py`/`fieldnames`, please leave the last item in each list as the human readable description. The documentation webpage is build from these lists and assumes the final item to be the description. Thanks!

mommermi:
@mkelley Feel free to ignore the failed builds. This issue will be taken care of in #45

mommermi:
#45 is merged; let's see if that fixes the build issues. Restarted the build of this PR.

mkelley:
Thanks.  Will get back to this after ADASS.

mommermi:
#45 is merged and the build succeeds; is this ready to be merged now?

mkelley:
I'm on it.

So I will make an exception for all dRA columns from Horizons and make them dRA cos(Dec).  Should we fix `astroquery`, too?

mommermi:
Thanks, no rush!

I think we should be consistent, so yes, probably we should change the column name in `jplhorizons`, too. I can take care of this. There is an open issue I have to take care of anyway.

mkelley:
Thanks, also check out my PR astropy/astroquery#1284

mkelley:
How do you feel about 'RA*cos(dec)_rate' as a column name?  jplhorizons returns RA_rate and Dec_rate so trying to find a name that preserves symmetry.

mommermi:
... it gives me a heart-burn - but I think it makes sense. Go for it!

mkelley:
Hmmm, looks like the same issue with oorb: `u.isclose(horizons_ephem['ra_rate'][0], oo_ephem['ra_rate'][0]`

mkelley:
pyoorb seems to only return dRA*cos(Dec), too.  I'll make a similar fix for it too.

mommermi:
yessir, thank you!

mkelley:
OK, it crashed trying to import the ginga_plugins submodule, but I think it just did this for fun, because AFAIK nothing tests or imports it.  How should I address this?

PR #93, opened 2018-12-04
Better handling of optional packages in ginga_plugins.

mkelley:
Optional packages are:
* ginga
* photutils

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -40,6 +40,20 @@ your default Python (this will be asked during the installation process).
 Requirements
 ^^^^^^^^^^^^
 
+`sbpy` has the following requirements (incomplete):
+
+* Python 3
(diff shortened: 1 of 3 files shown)


mkelley:
Looks like this is working, @gianninapr .  @mommermi or @migueldvb , should this be all I need to do?

mkelley:
Let me add some notes in the documentation.

gianninapr:
@mkelley I'll see if the build of PR #92 runs without ginga as a requirement, if it does then that should be good, at least from my part

mkelley:
I put a space in the main documentation page for lists of required and optional packages.  I know this is incomplete, and I'm sure that it is evolving.  Unless something is totally wrong in this PR, I recommend address fixes in another issue / PR.

mommermi:
looks good to me, thanks for taking care of this!

PR #95, opened 2018-12-07
Updating astropy_helpers to 2.0.8

migueldvb:
This merges the latest astropy-helpers 2.0.8. Hopefully all the checks will pass.


mkelley:
Since we're no longer supporting python2, should this be a 3.x branch of AH?

mkelley:
Ah, or do we need to match the earliest astropy that we support, which I think is still 2.x.

migueldvb:
I think that is true. We could stay with astropy-helpers 2.0.8 for the time being.

mkelley:
Sounds good.

Issue #91, opened 2018-11-18
Ephem.from_oo modifies orbit parameter in place


mkelley:
``Ephem.from_oo`` modifies the ``orbit`` parameter in place.  Then when reusing the orbit, ``from_oo`` crashes because the column ``'orbittype'`` already exists.

(1) Consider copying the orbit object before modifying.
(2) Check for existence of ``'orbittype'`` column before adding it.
```
In [15]: orbit = Orbit.from_dict({
    ...:     'targetname': ['A109AGz'],
    ...:     'a': [26.0446738] * u.au,
    ...:     'e': [0.9049629],
    ...:     'i': [159.11197] * u.deg,
    ...:     'Omega': [356.78024] * u.deg,
    ...:     'w': [318.64862] * u.deg,
    ...:     'M': [359.78524] * u.deg,
    ...:     'epoch': [Time('2018-10-29').jd],
    ...:     'timescale': ['UTC'],
    ...:     'H': [16.1],
    ...:     'G': [0.15]
    ...: })

In [16]: eph = Ephem.from_oo(orbit, epochs=Time('2018-11-18 21:00').jd + np.arange(25))

In [17]: eph = Ephem.from_oo(orbit, epochs=Time('2018-11-18 21:00').jd + np.arange(25))
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-17-d45a0fe59320> in <module>()
----> 1 eph = Ephem.from_oo(orbit, epochs=Time('2018-11-18 21:00').jd + np.arange(25))

/home/msk/local/lib/python3.6/site-packages/sbpy-0.0.dev362-py3.6.egg/sbpy/data/ephem.py in from_oo(self, orbit, epochs, location, scope, timescale, dynmodel, ephfile)
    529 
    530         # add orbittype column
--> 531         orbit.add_column([orbittype] * len(orbit.table), name='orbittype')
    532 
    533         # derive and apply default units

/home/msk/local/lib/python3.6/site-packages/sbpy-0.0.dev362-py3.6.egg/sbpy/data/core.py in add_column(self, data, name, **kwargs)
    538         """
    539 
--> 540         self._table.add_column(Column(data, name=name), **kwargs)
    541         return len(self.column_names)
    542 

/home/msk/local/lib/python3.6/site-packages/astropy-3.0.3-py3.6-linux-x86_64.egg/astropy/table/table.py in add_column(self, col, index, name, rename_duplicate, copy)
[...]

PR #101, opened 2019-01-18
fixes to data module

mommermi:
* fix for #91: the object in place is now not being modified
* added better handling for `__getitem__` in `DataClass`: if several columns are queried, a new `DataClass` object is returned, not just a table
* added some tests and documentation
* reflect changes to `pyoorb` in `Ephem` and `Orbit`


mommermi:
merge since built fails are unrelated

PR #103, opened 2019-01-20
SpectralGradient

mkelley:
Photometry seems like the right place to me.  I repeated and improved code for flux/mag conversions in `Afrho`, and `Efrho`.  I'd like to sort that out into a single shared code base before finishing this.


mkelley:
Notes: alpha = (1 + S * dlam / 2) / (1 - S * dlam / 2)

mkelley:
Wait for #117 before merging.

mkelley:
@jianyangli , I changed the units import in photometry to `import astropy.units as u` which is a common convention.  If this is a problem, let me know and I will change it back.

jianyangli:
@mkelley I already noticed it, and not a problem at all.  I'll review this PR later.

sbpy/spectroscopy/core.py
```diff
@@ -1046,3 +1047,51 @@ def filt(self, bp, unit='W / (m2 um)', **kwargs):
             fluxd = obs.effstim(_unit)
 
         return wave, fluxd
+
+    def color_index(self, wfb, unit):
```
jianyangli: Should we allow additional equivalencies to be accepted in case this class needs to be used for customized magnitude system?

`def color_index(self, wfb, unit, equivalencies=None):`

sbpy/photometry/core.py
```diff
@@ -99,24 +104,306 @@ def mag2ref(mag, radius, M_sun=None):
...
+            S.wave0 = wave0
+
+        return S
+
+    @classmethod
+    def _eff_wave(cls, wfb):
+        """Wavelength/frequency/bandpass to wavelength.
+
+        Bandpass is converted to effective wavelength using a solar
+        spectrum.
```
jianyangli: Should the option be open for a reference spectrum other than solar when calculating effective wavelength from bandpass?  Might be useful for extremely red object, but I don't know how of difference might be, and therefore the necessity.  Or maybe we can redden the solar spectrum based on `S.value` for effective wavelength calculation?

sbpy/photometry/core.py
```diff
@@ -99,24 +104,306 @@ def mag2ref(mag, radius, M_sun=None):
...
+
+        Computes spectral gradient from ``color_index``.
+        ``wfb[0]`` is the blue-ward of the two measurements
+
+        .. math::
+
+           S &= \frac{R(λ1) - R(λ0)}{R(λ1) + R(λ0)} \frac{2}{Δλ} \\
+           &= \frac{α - 1}{α + 1} \frac{2}{Δλ}
+
+        where R(λ) is the reflectivity, and:
```
jianyangli: reflectivity or reflectance?  Have to admit that I don't really know the difference between these two words.

sbpy/photometry/core.py
```diff
@@ -99,24 +104,306 @@ def mag2ref(mag, radius, M_sun=None):
...
+        beta = (S * dw / 2).decompose()  # dimensionless
+        color = ((1 + beta) / (1 - beta)).to(u.mag, u.logarithmic())
+
+        return color
+
+    def renormalize(self, wave0):
+        """Re-normalize to another wavelength.
+
+        The slope is linearly extrapolated to the new normalization
+        point.  Requires the `wave` attribute to be defined, see
```
jianyangli: `wave0` is required, but `wave` isn't, right?

sbpy/spectroscopy/core.py
mkelley: Yes, that is the right approach.

sbpy/photometry/core.py
mkelley: Oh, that'll be worth investigation to see how large an effect it could be.  I'm fine with an iteration, but it would have to assume a functional form of the reddening (probably linear).  Maybe an optional iteration?

sbpy/photometry/core.py
mkelley: Correct, will fix that too.

sbpy/photometry/core.py
jianyangli: Optional iteration is fine because it has the risk of slow and not converge.   Linear reddening is the only option we can take I guess.

sbpy/spectroscopy/core.py
mkelley: I don't have any tests in this PR, so I'd rather address this in #124 .

sbpy/photometry/core.py
mkelley: We should make a reddening function, or look into synphot for what is used there.  Since we agree that this should be optional, I recommend we save it for v0.2.

jianyangli: Sounds good to me.

sbpy/photometry/core.py
mkelley: I used reflectivity here since that is what Mike A. used in his paper.  I think they are interchangable, though.

mkelley:
Thanks for the review, @jianyangli .

Issue #100, opened 2018-12-21
Ephem.from_horizons can't handle multiple Time values


jianyangli:
Below code generates an error message:
```
from sbpy.data import Ephem
from astropy.time import Time
epochs = Time(['2015-01-01', '2015-01-02'])
eph = Ephem.from_horizons('Ceres', epochs=epochs)
```

The problem is that `.from_horizons` doesn't handle `epochs` value correctly when it contains multiple values.  There are also other bugs for the case `isinstance(epochs, (list, tuple, ndarray))`.

jianyangli:
I think we should either describe what data type it accepts clearly in the docs, or make it truly flexible. 
But I just noticed that Michael did note it in the code that this input needs to be modified.  So the potential problems here are in his radar already.

jianyangli:
I'm closing this issue now as the problem has been resolved with PR #106

PR #106, opened 2019-01-30
Fix for Issue #100: from_horizons now supports multiple Time values

gianninapr:
This fixes issue #100 to support all instances that are allowed according to the documentation: None, Time object, Time object which date values are a list of dates, dictionaries, plain list of julian dates, and a list of separately created Time objects. If we prefer shorter implementation code we could tweak the documentation to only allow None, dictionaries, or Time objects with either 1 date or a list of dates as such:

`epochs = Time('2015-01-01')`

or

`epochs = Time(['2015-01-01', '2015-01-02'])`

Let me know if there's any advice for anything to add or if we want to limit the user options for the input in order to make the code more concise.

### DIFF
--- a/sbpy/data/ephem.py
+++ b/sbpy/data/ephem.py
@@ -79,7 +79,13 @@ class Ephem(DataClass):
         if epochs is None:
             epochs = [Time.now().jd]
         elif isinstance(epochs, Time):
-            epochs = [Time(epochs).jd]
+            epochs = epochs.jd
+            if isinstance(epochs, float):
+                epochs = [epochs]
+            new_epochs = [None] * len(epochs)
+            for i in range(len(epochs)):
+                new_epochs[i] = epochs[i]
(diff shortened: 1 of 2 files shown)


mkelley:
I think the master branch is back into shape.  @gianninapr , can you rebase?

I like limiting inputs to None, dictionaries, or Time.  What does @mommermi think?

mommermi:
Hmm. Forcing users to provide `... epoch={'jd':2154200}, ...`? At first I thought that would be a bad idea, but in the end, it's not that much more typing. And it might help educate users to actually prefer using `Time` objects... Let's discuss this on slack...

mommermi:
@gianninapr  this is ready to be merged, right?

gianninapr:
> @gianninapr this is ready to be merged, right?

Yes

mommermi:
Perfect, thanks, Giannina!

PR #107, opened 2019-02-04
minor fixes to openorb functionality

mommermi:
output angles now in degrees; output epochs in JD instead of MJD


mommermi:
Travis-CI will not trigger build using latest commit - close this PR and reopen

Issue #116, opened 2019-03-01
synphot version requirement or warning


mkelley:
There are some percent-level and larger errors in synphot, to be fixed with spacetelescope/synphot_refactor#166 .  Fortunately, I don't think it affects sbpy tests at anything greater than the percent level.  We should probably make a version requirement or throw a warning if an older version is used.

mkelley:
@mommermi How are package requirements specified?  Let's discuss tomorrow.

mkelley:
We've been pinned to synphot>=1.0 via #284

PR #114, opened 2019-02-26
Magnitude calibration

mkelley:
Calibration scheme for magnitude and flux density conversions.

1. Adds a new unit module, and a unit named `VegaMag`.  This is appropriately defined using astropy's magnitude units, similar to ABmag and STmag.  I thought about just importing synphot's VEGAMAG, but that one does not use astropy's magnitude units.

2. New photometry.calibration with `fluxd_to_mag` and `mag_to_fluxd`.  `fluxd_to_mag` is tested against Willmer 2018's Vega and Sun values for the SDSS r filter.  Agreement is within 0.01 mag.

3. `mag_to_fluxd` tests coming soon, but will just be the opposite of 2.

The PR will help me avoid some repeated code I had in Afrho, Efrho, and my new SpectralGradient class.  I think it may be useful for some of the photometry methods, and if we do any H-mag to flux density conversions.

PR #117, opened 2019-03-02
New units equivalency for Vega magnitudes.

mkelley:
Conversion to/from Vega magnitudes with a units equivalency function in a new `units` submodule.  Uses `VEGAMAG` from `synphot` and imports it for convenience:

```python
  >>> import astropy.units as u
  >>> from sbpy.units import VEGAMAG, spectral_density_vega
  >>> m = 0.03 * VEGAMAG
  >>> fluxd = m.to('erg/(cm2 s AA)', spectral_density_vega(5500 * u.AA))
  >>> fluxd.value                                  # doctest: +FLOAT_CMP
  3.5469235179497687e-09
  >>> m = fluxd.to(VEGAMAG, spectral_density_vega(5500 * u.AA))
  >>> m.value                                      # doctest: +FLOAT_CMP
  0.03
```

Conversion also can use a frequency (not shown) or bandpass (via `synphot`):
```python
  >>> from sbpy.units import VEGAMAG, spectral_density_vega
  >>> m = 0.03 * VEGAMAG
  >>> fluxd = m.to('erg/(cm2 s AA)',
  ...     spectral_density_vega('johnson_v'))  # doctest: +REMOTE_DATA
  >>> fluxd.value                   # doctest: +FLOAT_CMP +REMOTE_DATA
  3.588524658721229e-09
```

Tests are good at the 0.5 to 1.4% level.  Perhaps they will improve once `synphot` is updated (re: #116)?


mkelley:
This replaces #114.  Sun calibration tests will be updated in a future PR.

mommermi:
@mkelley just a thought: might the `astropy.units` developers be interested in putting this as an equivalency into their package?

mkelley:
Pey-Lian thinks it is too complicated for astropy.  synphot handles it well enough, but I wanted more direct support for bandpasses in sbpy, hence the PR.

mommermi:
Ok, sounds good to me. I was just wondering.

jianyangli:
To verify the conversion between flux density and Vega mag, I took the PHOTFLAM values listed in the HST/WFC3/UVIS [photometric zeropoint table](http://www.stsci.edu/hst/wfc3/analysis/uvis_zpts/uvis2_infinite) for UVIS2 chip at infinite aperture, and calculated the corresponding Vega mags for some filters with the transmission curves downloaded from [HST CDBS website](ftp://ftp.stsci.edu/cdbs/comp/wfc3/), and compared with the two Vega mag values listed in the zeropoint table.  I also calculated the solar magnitude through these filters using `sbpy.spectroscopy.sun.Sun` to compare with the values that I calculated earlier.  Results are shown in the figure below.

image

Results:
1. The difference between the two Vega mag listed in the zeropoint table are close to each other except for the UV filters.
1. The difference for all wide band filters centered between F336W and F775W, and all medium band and narrowband filters are within 0.02 mag.
1. The differences are larger and could be up to 0.38 mag for long-pass filters.
1. The UV filters also have some discrepancies up to 0.18 mag.
1. There seems to be some systematic trend from UV to NIR for wide band filters.
1. The bias in PHOTFLAM Vega mag is almost the same as that in solar mag
1. The omit of the transmission curves of other optic elements in my calculation cannot be excluded as a possible cause of the bias.

Conclusion:
1. Overall the conversion algorithm seems to work well.
1. There seems to be a systematic bias of about 0.02 mag, as suggested by the bias of narrowband and medium band filters, and wide band filters between F336W and F625W
1. The consistent bias between PHOTFLAM Vega mag and solar mag suggests that the cause of the ~0.2 mag bias is in the unit conversion, not in transmission curves or others.

I see that in function `sbpy.units.spectral_density_vega()`, the Vega flux standard is multiplied by a factor of 1.0280163.  Maybe this is the cause of ~0.02 mag higher magnitude compared to the HST zeropoint table?

jianyangli:
OK, seems like HST is assuming 0 magnitude for Vega, instead of 0.03.  See [http://www.stsci.edu/hst/wfc3/phot_zp_lbn](http://www.stsci.edu/hst/wfc3/phot_zp_lbn) near the bottom of the page.  So what's the standard?  @mkelley

mkelley:
Great idea to use this table to test the code.

Regarding 0.03 vs. 0.00, Bohlin (2014) is certainly aware that V=0.03 for Vega, but ACS, WFC3, and Gaia all take the VEGAMAG system to mean m_Vega = 0.00.

Maybe I'm confused about the nomenclature.  If the VEGAMAG system assumes 0.00, then when would we use the 0.03 offset?  Bessell & Murphy (2012) seem to call this the V system or UBV system.  Wikipedia has UBV system, or Johnson system, or Johnson-Morgan system.  Landolt (1983, 1992, 2009) uses the UBV system, but I haven't found a discussion of zeropoints for his papers, other than references to Johnson and Morgan (1953) and similar.  Johnson et al. (1966, UBVRIJKL Photometry of the Bright Stars) reports Vega with V=0.03 mag.

OK, so my suggestion is to adopt 0.00 at all wavelengths for VEGAMAG, and to define a second equivalency that assumes V=0.03.  How about ``spectral_density_johnson66`` or ``spectral_density_ubvsys``?

I'll look into the other issue later.

mkelley:
I think better than a second equivalency is a new magnitude unit, ``ubvmag`` or ``j66mag``, and ``spectral_density_vega`` would use the 0.03 mag offset as needed.

jianyangli:
I'm confused here, too.  Also I thought Vega magnitude should be the same in any wavelength, or any bands, so it's independent of the filter system.  The name `ubvmag` seems confusing to me (or I could be wrong in understanding the Vega mag system).

If we don't want to deal with this, a workaround might be to send a keyword to the equivalency definition to define the magnitude of Vega?

```
def spectral_density_vega(wfb, vegamag=0.03):
   ...
```

jianyangli:
I'm actually not too concerned by the discrepancies in those long-pass filters, or the apparent trend in the bias with respect to wavelength for wide band filters.  Those could be due easily to the different Vega spectrum, solar spectrum, or throughput curves (which could be interacting with the CCD QE curve near the boundary of the response) that I used from what HST uses.  So I would suggest that once we deal with the Vega mag definition, this branch will be ready to deploy.

mkelley:
Yeah, ubvmag is a confusing name.  This is a zero-point flux density issue, and zero-points are tied to the units behind the magnitude.  I think best is a separate magnitude unit.  I still like ``j66mag``.

mkelley:
@jianyangli I've updated the code.  Now I have defined a new unit called 'Vega', upon which the magnitude systems are defined.  This is slightly more sophisticated than `VEGAMAG` from `synphot`, so I have dropped support for that unit to keep the code simple.  Let me know if you think it is important to restore it.  I you use `sbpy.units.vega_mag` and redo your conversions, they should have a better agreement with the table.

mkelley:
Cleaned up the code a bit, aiming to make the unit names consistent with astropy, which, e.g., uses AB for the zeropoint of the AB magnitude system:
```python
>>> import astropy.units as u
>>> import sbpy.units
>>> sbpy.units.enable()
>>> m1 = u.Magnitude(0, 'mag(VEGA)')
>>> m2 = u.Magnitude(0, 'mag(JM)')
```
where JM represents the Johnson-Morgan system (Vega == 0.03 mag).

mkelley:
Checking that HST table again, it says, "... calculated from the master drizzled images and an improved fit to the observed to synthetic photometry."  So I agree with you that the discrepancies for the wide filters are probably due to some effects not accounted for (or cannot be accounted for) in the bandpass files.

docs/sbpy/units.rst
```diff
@@ -0,0 +1,61 @@
...
+`~sbpy.units` provides common planetary astronomy units and Vega-based
+magnitude conversions.  ``sbpy`` units may be added to the top-level
+``astropy.units`` namespace via:
+
+  >>> import sbpy.units
+  >>> sbpy.units.enable()    # doctest: +SKIP
+
+This will make them searchable via strings, such as
+``u.Unit('mag(VEGA)')``.
+
```
jianyangli: Just a heads-up, later I will add some reflectance units such as I/F, RADF, REFF, etc to this submodule, and update the description here.

jianyangli:
@mkelley I like the new implementation.  I think it has a better integration with `astropy`.  I tested the new unit and now the difference between what I calculated and what's in the table are less than 0.01 mag for PHOTFLAM to VEGAMAG conversion and for solar magnitude for all the wide band filters between F336W and F625W and all narrow band filters.  I think this branch is now ready to deploy.

image

docs/sbpy/units.rst
mkelley: I also have a ``hundred_nm`` unit to add for spectral gradients.

mkelley:
Thanks!  Merging...

Issue #122, opened 2019-03-19
Finalize SpectralStandard


mkelley:
There is a philosophical issue with the current state of `SpectralStandard`.  `__call__` will interpolate if a single wavelength is passed, but will rebin when multiple wavelengths are passed. 
Instead, I propose to only interpolate with `__call__` and make a new method, `observe()` which will always rebin.  It could also handle multiple bandpasses, effectively replacing `filt()`.

```python
sun = sbpy.spectroscopy.sun.Sun()
w = [0.4, 0.5, 0.6] * u.um

# interpolated values
sun(w)

# rebinned values, i.e., return three flux densities centered on 0.4, 0.5,
# and 0.6 μm, averaged over 0.1 μm wide bandpasses
sun.observe(w)

# observe through an array of filters
sun.observe(('johnson_b', 'johnson_v', 'cousins_r'))
```

Also, I need to clean up the documentation and imports.  Complete tests.

mkelley:
Done!

PR #124, opened 2019-03-21
Observe standards

mkelley:
Addresses #122, based on #103.

* New behavior for `SpectralStandard.__call__()` - now it only interpolates (never rebin).
* New `SpectralStandard.observe()` - for all your rebinning needs, including observing through a list of filters.
* Added Johnson V bandpass to sbpy files: it is too fundamental for the tests to leave as a remote dependency.
* New `Sun.show_builtin()` and `Vega.show_builtin()` to list available spectra.
* Revised tests for 100% coverage on SpectralStandard, Sun, and Vega.
* Revised spectroscopy tests.

```python
>>> from sbpy.spectroscopy import Sun
>>> Sun.show_builtin()
    name                                description                           
------------ -----------------------------------------------------------------
   E490_2014                E490-00a (2014) reference solar spectrum (Table 3)
 E490_2014LR E490-00a (2014) low resolution reference solar spectrum (Table 4)
  Kurucz1993               Kurucz (1993) model, scaled by Colina et al. (1996)
Castelli1996      Castelli model, scaled and presented by Colina et al. (1996)
```


mkelley:
Add `equivalencies` option to `color_index`.

mkelley:
@jianyangli I add the equivalencies option to `color_index`, as requested.  However, I did this as two parameters: (1) the `equivalencies` keyword argument, which works just like in `Unit.to`; (2) the `equiv_func` keyword argument, which takes a function that will return a list of equivalencies.  That way something like `spectral_density_vega` can be passed and used for each calculated effective wavelength.  I'm open to another solution, though.  What do you think?

mkelley:
@jianyangli Besides changing the keyword name from `equiv_func`, I haven't come up with another approach.  I looked through the list of astropy affiliated packages, and I didn't see anything that might serve as an example.

jianyangli:
@mkelley Sorry again for my delay replying you - got tied up with something else last week.  This approach looks good to me.  If we want to make the interface more concise, maybe we just make `equivalencies` more flexible and allow it to take either a list of equivalencies or an equivalency function?  I can't think of a case when both a list and a function are needed at the same time.  But your current approach is probably more conceptually clean.  I'm fine with either way.

mkelley:
I thought of that too, and actually started out that way, but in the end favored two separate keywords as it made the code simpler.  Thanks.

PR #128, opened 2019-03-27
Magnitude-reflectance-cross-section equivalencies

jianyangli:
Added unit conversion equivalencies between magnitude and reflectance for given scattering cross-section, and between magnitude and scattering cross-section for given reflectance.

This PR can wait until after the v0.1 release.

### DIFF
--- a/sbpy/units/core.py
+++ b/sbpy/units/core.py
@@ -17,7 +17,8 @@ __all__ = [
     'VEGA',
     'VEGAmag',
     'JM',
-    'JMmag'
+    'JMmag',
+    'magnitude_reflectance'
 ]
 
 from warnings import warn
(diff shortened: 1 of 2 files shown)


sbpy/units/core.py
```diff
@@ -125,3 +128,148 @@ def spectral_density_vega(wfb):
...
+
+    Parameters
+    ----------
+    ref : `~astropy.units.Quantity`
+        Average reflectance of the object of interest
+    wfb : `~astropy.units.Quantity`, `~synphot.SpectralElement`, string
+        Wavelength, frequency, or a bandpass of the corresponding flux
+        density being converted.  See
+        :func:`~synphot.SpectralElement.from_filter()` for possible
+        bandpass names.  If provided, then this parameter overrides `M_sun`.
```
mkelley: What should be the behavior for multiple wavelengths?  I think it should use the `Sun.observe` method in #124, which would treat a set of wavelengths as the bin centers of a spectrum, rather than interpolate the solar spectrum to each wavelength.  Whichever the case, the assumption should be documented in the docstring.

sbpy/units/tests/test_core.py
```diff
@@ -90,3 +90,33 @@ def test_spectral_density_vega_bp(filename, fluxd, to, tol):
...
+
+
+@pytest.mark.parametrize('mag, wfb, M_sun, ref', (
+    (3.4 * VEGAmag, None, None, 0.02865984),
+    (3.4 * VEGAmag, 'johnson_v', None, 0.02865984),
+    (3.4 * VEGAmag, 'johnson_i', None, 0.01387486),
+    (3.4 * VEGAmag, 5500 * u.AA, None, 0.02774623),
+    (3.4 * VEGAmag, None, -25.0 * VEGAmag, 0.14694669),
+    (3.4 * u.ABmag, None, -25.0 * u.ABmag, 0.14694669),
+))
```
mkelley: Are there any literature values that we can test against?

sbpy/units/core.py
```diff
@@ -125,3 +128,148 @@ def spectral_density_vega(wfb):
         (flam0.unit, VEGA, lambda x: x / flam0.value,
          lambda x: x * flam0.value)
     ]
+
+
+def magnitude_reflectance(xsec, wfb=None, M_sun=None):
```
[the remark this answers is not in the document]
mkelley: For example, if I want to convert Deep Impact spectra to reflectance, I use a function named `magnitude_reflectance`, which does not seem right.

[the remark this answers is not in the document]
jianyangli: You are right here, but in the future I will define some equivalencies to support conversion between various reflectance units (bidirectional, RADF, REFF, BRDF, etc).  I think it's more appropriate to call that equivalency `reflectance`.  What about we call this one `flux_reflectance`?

sbpy/units/core.py
jianyangli: Agreed.  Will wait for #124 to be merged before finalize this PR.

sbpy/units/tests/test_core.py
jianyangli: Unfortunately I don't think so.  The best would be to use some  objects, such as Ceres, for which we have good measurements for size, magnitude, and reflectance.  The test for `M_sun=-25` and for `ABmag` is just to verify the code works, but there should be a better way.  Will revise it.

sbpy/units/core.py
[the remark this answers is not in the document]
mkelley: I looked at http://docs.astropy.org/en/stable/units/equivalencies.html for guidance, and equivalency names are more about the concept rather than the two items being converted.

For all these reflectance units, can they not just be returned by equivalency, like the way that `u.spectral_density` returns 16 different conversions?  Why separate functions for each one?

sbpy/units/tests/test_core.py
mkelley: Using Ceres sounds like a great test.

sbpy/units/core.py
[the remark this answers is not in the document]
jianyangli: Good point.  We can do that.

sbpy/units/tests/test_core.py
jianyangli: Well, that's what I've been doing with all the `3.4 * VEGAmag` :)

sbpy/units/core.py
```diff
@@ -125,3 +128,148 @@ def spectral_density_vega(wfb):
...
+        if not hasattr(M_sun.unit, 'physical_unit'):
+            raise u.UnitTypeError('the magnitude system must be based on a physical unit')
+        f_sun = M_sun.to(M_sun.unit.physical_unit).value
+    return [(M_sun.unit.physical_unit,
+             u.sr**-1,
+             lambda mag: mag/(f_sun*xsec.to('au2').value),
+             lambda ref: ref*f_sun*xsec.to('au2').value)]
+
+
+def magnitude_xsection(ref, wfb=None, M_sun=None):
```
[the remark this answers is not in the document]
mkelley: Actually. I prefer `cross_section` over `xsection`, slightly easier for me to remember and for a new user to discover.  But it isn't a big deal.

jianyangli:
Sorry Mike, click request review by mistake.  Haven't done anything yet.  Please ignore this request.

sbpy/units/core.py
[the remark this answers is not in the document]
jianyangli: I merged the two conversions into one single equivalence function `reflectance`, and will use this same function to support conversions between different reflectance units in the future.

sbpy/units/core.py
```diff
@@ -125,3 +128,148 @@ def spectral_density_vega(wfb):
         (flam0.unit, VEGA, lambda x: x / flam0.value,
          lambda x: x * flam0.value)
     ]
+
+
+def magnitude_reflectance(xsec, wfb=None, M_sun=None):
+    """Magnitude - reflectance equivalencies to convert between magnitude and
+    average reflectance for given scattering cross-section.
```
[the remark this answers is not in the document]
jianyangli: Done

sbpy/units/core.py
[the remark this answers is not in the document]
jianyangli: Use `cross_section` in all occurrences.

sbpy/units/core.py
```diff
@@ -125,3 +128,148 @@ def spectral_density_vega(wfb):
...
+        f_sun = M_sun.to(M_sun.unit.physical_unit).value
+    return [(M_sun.unit.physical_unit,
+             u.sr**-1,
+             lambda mag: mag/(f_sun*xsec.to('au2').value),
+             lambda ref: ref*f_sun*xsec.to('au2').value)]
+
+
+def magnitude_xsection(ref, wfb=None, M_sun=None):
+    """Magnitude - cross-section equivalencies to convert between magnitude
+    and scattering cross-section for given average reflectance.
```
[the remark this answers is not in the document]
jianyangli: N/A any more.

sbpy/units/core.py
jianyangli: @mkelley I admit I'm confused here.  Why would we need to worry about the case of multiple wavelength?  The quantity to be converted is always observed at one single wavelength or band, right?  Furthermore, when multiple wavelengths/bands are passed, `Sun.observe` would return multiple flux density values.  How would we use multiple values, average them?

mkelley: I was thinking about the case of a low-resolution spectrum.  Each element in the spectrum is the average flux density over a small wavelength range.  The solar spectrum should also be averaged over the small wavelength range.  Interpolation to the central wavelength could be wrong.  This matters most near absorption features a higher resolution solar spectra.

mkelley: With a single wavelength, all we can ever do is interpolate the solar spectrum and the user must accept the possibility of a bias.  In such a case a bandpass would be more appropriate.

Back to the spectrum example.  If you want to convert a Deep Impact spectrum of the surface of Hartley 2 to a reflectance spectrum, I think it should be allowed to pass all wavelengths at once, rather than iterating through each element.

sbpy/units/core.py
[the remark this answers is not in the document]
mkelley: Sounds good.

sbpy/units/core.py
jianyangli: OK I got it.  Do you have any good spectral data for testing?

mkelley: How about Hartley 2?

mkelley: Tempel 1 will be easier.

jianyangli:
Closed and replaced by #136

PR #127, opened 2019-03-24
Photometry

jianyangli:
The PR is to support release v0.1.  The main updates on this PR are:

1. Implemented the revised HG12 model by Penttilä et al. (2016).
1. Add `.from_data()` class method
1. Add `.fit()` method as a wrapper to take `sbpy.data.Ephem` as input
1. Update the input/output to a few methods to allow `sbpy.data.Ephem`
1. Add support for units in data and models
1. Other miscellaneous update to support the above changes
1. Miscellaneous bugfix

This PR is also related to issue #112

PR #134, opened 2019-04-09
Photometry rebased

jianyangli:
Rebased on master with squash merge to replace PR #127 .

commit 7f1556c302c2cca8af8b0234422dc2c57b2df842
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Sat Mar 30 22:29:53 2019 -0400

    add parameter validator tests

commit 2a905c00ad3b2c1c35d76eef90048e1569d33ec3
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Sat Mar 30 15:38:02 2019 -0400

    add `validator` for model parameters

commit 3adf5f20fb31d678b6a47b39bbeb7440b6da08ca
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Sat Mar 30 10:44:24 2019 -0400

    remove `test_spline`

commit 76378cca8c4746ab8572c2608586eb3fa35bd632
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Wed Mar 27 09:21:31 2019 -0400

    make `spline` private

commit 1c003c2a3fd28460a08eafa57e2ff733e440b71f
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Tue Mar 26 21:34:21 2019 -0400

    warnings for possible non-monotonic phase function

commit 4f2000808ec3b65f3131e3a7100804eecd3d6f05
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Tue Mar 26 20:47:42 2019 -0400

    bugfix in mag unit of distance corrections

commit 54d4bfff745a7fb710cd9bbe8ea9273225e61bc6
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Sun Mar 24 21:44:39 2019 -0400

    fix tests

commit 1208e05ed96896647cf9941b627af22de3b9aea0
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Sun Mar 24 20:14:23 2019 -0400

    PEP8 compatibility

commit 55f6915aa583de62b836332baff0732e77724485
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Sun Mar 24 13:46:29 2019 -0400

    PEP8 compatibility

commit 1be790e40e390fc052fdb727efff40c0377574a8
Merge: 66c0173 726b272
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Sun Mar 24 12:24:46 2019 -0400

    Merge branch 'photometry' of https://github.com/jianyangli/sbpy into photometry

commit 66c0173c11c6df391217a866f95aad85d9264655
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Fri Mar 22 16:55:30 2019 -0400

    revise test and bugfix

commit 462fa9b2cfb1ece62bf0db2265cd76f7f0e55ff0
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Fri Mar 22 14:34:58 2019 -0400

    update test and bugfix

commit ae827edd89f6181c17f0f23213ca2dc07a02b6f1
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Fri Mar 22 14:18:34 2019 -0400

    bugfix

commit cd3176b433707276318627ef447e2ddca3b4399b
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Fri Mar 22 14:09:33 2019 -0400

    change output to comply with sbpy guideline

    Output changed for DiskIntegratedPhaseFunc.mag and .ref

commit 6e911761b711d491ea251df588aab34ea2a7534d
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Fri Mar 22 00:19:04 2019 -0400

    add _process_ephem_input()

    `_process_ephem_input()` as a consistent checking and processing facilities to provide flexible API that is consistent with `sbpy` API guideline.

commit 09c6c0eb430a6823c30374cc5b0b26b2d9986195
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Thu Mar 21 23:27:52 2019 -0400

    add docstrings, move `._fit` to `.fit`

commit 80f7ff19f4dcd0d0555a04eca412e1e8fdfdd32f
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Thu Mar 21 17:25:32 2019 -0400

    add unit support to disk-int phase function classes

commit 3314c2faa70341cd5b2a7dc0a84f3f88b2715323
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Thu Mar 21 15:49:44 2019 -0400

    change `DiskIntegratedModelClass` name

    …to DiskIntegratedPhaseFunc

commit 563a5461893bd24d86bacc048c957bf8f95d874f
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Wed Mar 20 23:11:32 2019 -0400

    add `.from_data` class method

    1. Class method `.from_date` added to disk-integrated phase function model class to allow instantiate model class direction from data through fitting
    2. Implemented `_distance_module()` to calculate distance correction.
    3. Add default values for IAU HG, H-G1-G2, HG12, and HG12_Pen models
    4. Remove redundent code for HG12_Pen
    5. Updated test code to accomodate commit fd94056

commit 82dca9956183f223895d3774ddbdb06365bda945
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Wed Mar 20 09:44:44 2019 -0400

    change `astropy.units` to `u`

commit 2c75875524dbbefdffa1a4c9a3b4e9c833a6bc28
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Thu Feb 7 22:16:36 2019 -0500

    implemented the revised HG12 model by Penttila et al. 2016

commit 1dc4f41b78e5e9edae5adad45ca87921a92b7950
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Wed Dec 19 11:35:37 2018 -0500

    bugfix

    This bug results in wrong calculations in `HG12.fit_deriv()`

commit 726b2723819e33793ab5a3c28696c5c9597932a6
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Fri Mar 22 16:55:30 2019 -0400

    revise test and bugfix

commit 148b75e7824065b4923fb5efc860cb909b450381
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Fri Mar 22 14:34:58 2019 -0400

    update test and bugfix

commit 68c02639ae51acd791eefba927e6b04e1a8f86bf
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Fri Mar 22 14:18:34 2019 -0400

    bugfix

commit 934721300cbbbf54f3e7af68e6999ec17b855249
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Fri Mar 22 14:09:33 2019 -0400

    change output to comply with sbpy guideline

    Output changed for DiskIntegratedPhaseFunc.mag and .ref

commit faaf8ac80170518a3c37f9487789ee1223826c21
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Fri Mar 22 00:19:04 2019 -0400

    add _process_ephem_input()

    `_process_ephem_input()` as a consistent checking and processing facilities to provide flexible API that is consistent with `sbpy` API guideline.

commit 7619d1071ad2420863083bdf3cdbf1a7d1993837
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Thu Mar 21 23:27:52 2019 -0400

    add docstrings, move `._fit` to `.fit`

commit 6d05af54faf08dfe12d3b1b0d0c02e52d9dc5c83
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Thu Mar 21 17:25:32 2019 -0400

    add unit support to disk-int phase function classes

commit e5961f7e259692d0b8800f0e95db01ab66d79f44
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Thu Mar 21 15:49:44 2019 -0400

    change `DiskIntegratedModelClass` name

    …to DiskIntegratedPhaseFunc

commit 310c74c5456e08b42090c9798c58881c1e850ab2
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Wed Mar 20 23:11:32 2019 -0400

    add `.from_data` class method

    1. Class method `.from_date` added to disk-integrated phase function model class to allow instantiate model class direction from data through fitting
    2. Implemented `_distance_module()` to calculate distance correction.
    3. Add default values for IAU HG, H-G1-G2, HG12, and HG12_Pen models
    4. Remove redundent code for HG12_Pen
    5. Updated test code to accomodate commit fd94056

commit 5a6b832ab930e0781cd370b26c130e568dc6a8ae
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Wed Mar 20 09:44:44 2019 -0400

    change `astropy.units` to `u`

commit aa0d67632241e2747b348bee388f7c478c1c653a
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Thu Feb 7 22:16:36 2019 -0500

    implemented the revised HG12 model by Penttila et al. 2016

commit 0a8db401943407aa9b21d5ce2326ada3143339b4
Author: Jian-Yang Li <jianyang2li@gmail.com>
Date:   Wed Dec 19 11:35:37 2018 -0500

    bugfix

    This bug results in wrong calculations in `HG12.fit_deriv()`

### DIFF
--- a/sbpy/photometry/core.py
+++ b/sbpy/photometry/core.py
@@ -1288,6 +1288,13 @@ class HG1G2(HG12BaseClass):
 class HG12(HG12BaseClass):
     """HG12 photometric phase model (Muinonen et al. 2010)
 
+    This system is adopted by IAU as the "standard" model for disk-integrated
+    phase functions of planetary objects.  Note that there is a discontinuity
+    in the derivative for parameter G12, sometimes making the model fitting
+    difficult.  Penttil\"a et al. (2016, Planet. Space Sci. 123, 117-125)
+    revised the H, G12 system such that the G12 parameter has a continuous
+    derivative.  The revised model is implemented in class `G12_Pen16`.
+
     Examples
     --------
 
@@ -1374,6 +1381,12 @@ class HG12(HG12BaseClass):
 
 class HG12_Pen16(HG12):
     """Revised H, G12 model by Penttil\"a et al. (2016)
+
+    This system is the revised H, G12 system by Penttil\"a et al. (2016,
+    Planet. Space Sci. 123, 117-125) that has a continuous derivative with
+    respect to parameter G12.  The original model as adopted by IAU as the
+    "standard" model for disk-integrated phase functions of planetary objects
+    is implemented in class `HG12`.
 
     Examples
     --------
--- a/sbpy/photometry/core.py
+++ b/sbpy/photometry/core.py
@@ -54,8 +54,9 @@ def _process_ephem_input(eph, key=None):
     if eph is None:
         return None, None
     if not isinstance(eph, (Ephem, list, tuple, np.ndarray, Number,
-        u.Quantity)):
-        raise TypeError('`~sbpy.data.Ephem`, numbers, iterables of numbers, or `~astropy.units.Quantity` expected, {0} received'.format(type(eph)))
+                            u.Quantity)):
+        raise TypeError(
+            '`~sbpy.data.Ephem`, numbers, iterables of numbers, or `~astropy.units.Quantity` expected, {0} received'.format(type(eph)))
     if isinstance(eph, Ephem):
         if key is None:
             out = None
@@ -654,7 +655,8 @@ class DiskIntegratedPhaseFunc(Fittable1DModel):
                 for p in self.param_names:
                     par[p] = par[p][valid]
                 meta = kwargs.pop('meta', OrderedDict())
-                meta.update({'targetname': data['targetname'][valid]})
+                if 'targetname' in data.column_names:
+                    meta.update({'targetname': data['targetname'][valid]})
                 kwargs['meta'] = meta
                 for p in self.param_names:
                     val = kwargs.pop(p, None)
@@ -663,9 +665,14 @@ class DiskIntegratedPhaseFunc(Fittable1DModel):
                     radius = data['diameter'][valid]/2
                 except KeyError:
                     pass
-                log.info("Model initialized for {}.".format(meta['targetname']))
+                if 'targetname' in meta.keys():
+                    log.info("Model initialized for {}.".format(
+                        meta['targetname']))
+                else:
+                    log.info("Model initialized.")
             else:
-                raise ValueError('no valid model parameters found in `data` keyword')
+                raise ValueError(
+                    'no valid model parameters found in `data` keyword')
         super().__init__(*args, **kwargs)
         self.radius = radius
         self.M_sun = M_sun
@@ -1093,7 +1100,8 @@ class HG(DiskIntegratedPhaseFunc):
         and a warning will be issued.
         """
         if np.any(value > 1.194):
-            warnings.warn('G parameter could result in a non-monotonic phase function', RuntimeWarning)
+            warnings.warn(
+                'G parameter could result in a non-monotonic phase function', RuntimeWarning)
 
     @staticmethod
     def _hgphi(pha, i):
@@ -1250,7 +1258,8 @@ class HG1G2(HG12BaseClass):
         potentially be non-monotoic, and a warning will be issued.
         """
         if np.any(value < 0) or np.any(value + self.G2 > 1):
-            warnings.warn('G1, G2 parameter combination might result in a non-monotonic phase function', RuntimeWarning)
+            warnings.warn(
+                'G1, G2 parameter combination might result in a non-monotonic phase function', RuntimeWarning)
 
     @G2.validator
     def G2(self, value):
@@ -1260,7 +1269,8 @@ class HG1G2(HG12BaseClass):
         potentially be non-monotoic, and a warning will be issued.
         """
         if np.any(value < 0) or np.any(value + self.G1 > 1):
-            warnings.warn('G1, G2 parameter combination might result in a non-monotonic phase function', RuntimeWarning)
+            warnings.warn(
+                'G1, G2 parameter combination might result in a non-monotonic phase function', RuntimeWarning)
 
     @property
     def _G1(self):
@@ -1336,7 +1346,8 @@ class HG12(HG12BaseClass):
         non-monotoic, and a warning will be issued.
         """
         if np.any(value < -0.70) or np.any(value > 1.30):
-            warnings.warn('G12 parameter could result in a non-monotonic phase function', RuntimeWarning)
+            warnings.warn(
+                'G12 parameter could result in a non-monotonic phase function', RuntimeWarning)
 
     @property
     def _G1(self):
--- a/sbpy/photometry/core.py
+++ b/sbpy/photometry/core.py
@@ -36,7 +36,8 @@ def _process_ephem_input(eph, key=None):
 
     Parameters
     ----------
-    eph : `~sbpy.data.Ephem`, dict, ndarray, numbers, `~astropy.units.Quantity`
+    eph : `~sbpy.data.Ephem`, numbers, iterables of numbers, or
+            `~astropy.units.Quantity`
         The input to be processed.
     key : str
         The name of column to be extracted from `~sbpy.data.Ephem`.
@@ -44,25 +45,26 @@ def _process_ephem_input(eph, key=None):
     Returns
     -------
     eph : `~sbpy.data.Ephem`
-        If input `eph` can be converted to `~sbpy.data.Ephem`, then returns the
-        converted `~sbpy.data.Ephem` object.  If not, then returns `None`.
+        If `~sbpy.data.Ephem`, then returns itself.  If not, then returns
+        `None`.
     out : ndarray, numbers, `~astropy.units.Qauntity`
         The colume extracted from input `eph` if it can be converted to
         `~sbpy.data.Ephem`, or the original input `eph` if it cannot be converted.
     """
     if eph is None:
         return None, None
-    if not isinstance(eph, (Ephem, dict, np.ndarray, Number, u.Quantity)):
-        raise ValueError('`~sbpy.data.Ephem`, `dict`, `numpy.ndarray`, `Number`, or `~astropy.units.Quantity expected, {0} received'.format(type(eph)))
-    if isinstance(eph, dict):
-        eph = Ephem.from_dict(eph)
+    if not isinstance(eph, (Ephem, list, tuple, np.ndarray, Number,
+        u.Quantity)):
+        raise TypeError('`~sbpy.data.Ephem`, numbers, iterables of numbers, or `~astropy.units.Quantity` expected, {0} received'.format(type(eph)))
     if isinstance(eph, Ephem):
         if key is None:
             out = None
         else:
             out = eph[key]
     else:
-        out = eph
+        out = np.asanyarray(eph)
+        if not np.issubdtype(out.dtype, np.number):
+            raise TypeError('numerical type expected in iterables.')
         eph = None
     return eph, out
 
--- a/sbpy/photometry/core.py
+++ b/sbpy/photometry/core.py
@@ -1135,7 +1135,7 @@ class HG(DiskIntegratedPhaseFunc):
             func = func.value
         func = -2.5 * np.log10(func)
         if isinstance(hh, u.Quantity):
-            func = func * u.mag
+            func = func * hh.unit
         return hh + func
 
     @staticmethod
@@ -1277,7 +1277,7 @@ class HG1G2(HG12BaseClass):
             func = func.value
         func = -2.5 * np.log10(func)
         if isinstance(h, u.Quantity):
-            func = func * u.mag
+            func = func * h.unit
         return h + func
 
     @staticmethod
--- a/sbpy/photometry/core.py
+++ b/sbpy/photometry/core.py
@@ -61,6 +61,11 @@ def _process_ephem_input(eph, key=None):
             out = None
         else:
             out = eph[key]
+            if not isinstance(out, u.Quantity):
+                # When a column doesn't have a unit, ``out`` is not a
+                # `Quantity` but rather a `astropy.table.column.Column` object.
+                # This step is to ensure out as an array or Quantity.
+                out = out.data
     else:
         out = np.asanyarray(eph)
         if not np.issubdtype(out.dtype, np.number):
@@ -672,7 +677,7 @@ class DiskIntegratedPhaseFunc(Fittable1DModel):
     @property
     def geoalb(self):
         """Geometric albedo"""
-        alb = np.pi*self.ref(0)['ref'][0]
+        alb = np.pi*self.ref(0)
         if hasattr(alb, 'unit'):
             alb = alb*u.sr
         return alb
@@ -832,12 +837,13 @@ class DiskIntegratedPhaseFunc(Fittable1DModel):
             m0 = cls(*init)
         return m0.fit(eph, mag, fitter=fitter, **kwargs)
 
-    def mag(self, eph, **kwargs):
+    def mag(self, eph, append_results=False, **kwargs):
         """Calculate phase function in magnitude
 
         Parameters
         ----------
-        eph : `~sbpy.data.Ephem`, dict_like, float, or array_like of float
+        eph : `~sbpy.data.Ephem`, numbers, iterables of numbers, or
+            `~astropy.units.Quantity`
             If `~sbpy.data.Ephem` or dict_like, ephemerides of the object that
             can include phase angle, heliocentric and geocentric distances via
             keywords `phase`, `r` and `delta`.  If float or array_like, then
@@ -846,11 +852,21 @@ class DiskIntegratedPhaseFunc(Fittable1DModel):
             If no unit is provided via type `~astropy.units.Quantity`, then
             radians is assumed for phase angle, and au is assumed for
             distances.
+        append_results : bool
+            Controls the return of this method.
 
         Returns
         -------
-        `~sbpy.data.Ephem` object, with the calculated magnitudes added in
-        a new column.
+        `~astropy.units.Quantity`, array if ``append_results == False``
+        `~sbpy.data.Ephem` if ``append_results == True``
+
+        When ``apend_results == False``: The calculated magnitude will be
+        returned.
+
+        When ``append_results == True``:  If ``eph`` is a `~sbpy.data.Ephem`
+        object, then the calculated magnitude will be appended to ``eph`` as
+        a new column.  Otherwise a new `~sbpy.data.Ephem` object is created to
+        contain the input ``eph`` and the calculated magnitude in two columns.
 
         Examples
         --------
@@ -881,18 +897,27 @@ class DiskIntegratedPhaseFunc(Fittable1DModel):
             if isinstance(out, u.Quantity):
                 dist_corr = dist_corr*u.mag
             out = out - dist_corr
-        if eph is None:
-            eph = Ephem.from_dict({'alpha': pha, 'mag': out})
+        if append_results:
+            if eph is None:
+                eph = Ephem.from_dict({'alpha': pha, 'mag': out})
+            else:
+                name = 'mag'
+                i = 1
+                while name in eph.column_names:
+                    name = 'mag'+str(i)
+                    i += 1
+                eph.add_column(out, name=name)
+            return eph
         else:
-            eph.add_column(out, name='mag')
-        return eph
+            return out
 
-    def ref(self, eph, normalized=None, **kwargs):
+    def ref(self, eph, normalized=None, append_results=False, **kwargs):
         """Calculate phase function in average bidirectional reflectance
 
         Parameters
         ----------
-        eph : `~sbpy.data.Ephem`, dict_like, float, or array_like of float
+        eph : `~sbpy.data.Ephem`, numbers, iterables of numbers, or
+            `~astropy.units.Quantity`
             If `~sbpy.data.Ephem` or dict_like, ephemerides of the object that
             can include phase angle, heliocentric and geocentric distances via
             keywords `phase`, `r` and `delta`.  If float or array_like, then
@@ -901,11 +926,24 @@ class DiskIntegratedPhaseFunc(Fittable1DModel):
             If no unit is provided via type `~astropy.units.Quantity`, then
             radians is assumed for phase angle, and au is assumed for
             distances.
+        normalized : number, `~astropy.units.Quantity`
+            The angle to which the reflectance is normalized.
+        append_results : bool
+            Controls the return of this method.
 
         Returns
         -------
-        `~sbpy.data.Ephem` object, with the calculated reflectances added in a
-        new column.
+        `~astropy.units.Quantity`, array if ``append_results == False``
+        `~sbpy.data.Ephem` if ``append_results == True``
+
+        When ``apend_results == False``: The calculated reflectance will be
+        returned.
+
+        When ``append_results == True``:  If ``eph`` is a `~sbpy.data.Ephem`
+        object, then the calculated reflectance will be appended to ``eph`` as
+        a new column.  Otherwise a new `~sbpy.data.Ephem` object is created to
+        contain the input ``eph`` and the calculated reflectance in two
+        columns.
 
         Examples
         --------
@@ -938,11 +976,19 @@ class DiskIntegratedPhaseFunc(Fittable1DModel):
             out = mag2ref(out, self.radius, M_sun=self.M_sun)
             if normalized is not None:
                 out /= mag2ref(norm, self.radius, M_sun=self.M_sun)
-        if eph is None:
-            eph = Ephem.from_dict({'alpha': pha, 'ref': out})
+        if append_results:
+            if eph is None:
+                eph = Ephem.from_dict({'alpha': pha, 'ref': out})
+            else:
+                name = 'ref'
+                i = 1
+                while name in eph.column_names:
+                    name = 'ref'+str(i)
+                    i += 1
+                eph.add_column(out, name=name)
+            return eph
         else:
-            eph.add_column(out, name='ref')
-        return eph
+            return out
 
     def _phase_integral(self, integrator=quad):
         """Calculate phase integral with numerical integration
@@ -967,7 +1013,7 @@ class DiskIntegratedPhaseFunc(Fittable1DModel):
         0.364
 
         """
-        def integrand(x): return 2*self.ref(x, normalized=0.)['ref']*np.sin(x)
+        def integrand(x): return 2*self.ref(x, normalized=0.)*np.sin(x)
         return integrator(integrand, 0, np.pi)[0]
 
 
--- a/sbpy/photometry/core.py
+++ b/sbpy/photometry/core.py
@@ -893,7 +893,7 @@ class DiskIntegratedPhaseFunc(Fittable1DModel):
         >>> from sbpy.data import Ephem
         >>> ceres_hg = HG(3.34, 0.12)
         >>> # parameter `eph` as `~sbpy.data.Ephem` type
-        >>> eph = Ephem({'alpha': np.linspace(0,np.pi*0.9,200),
+        >>> eph = Ephem.from_dict({'alpha': np.linspace(0,np.pi*0.9,200),
         ...              'r': np.repeat(2.7*u.au, 200),
         ...              'delta': np.repeat(1.8*u.au, 200)})
         >>> mag1 = ceres_hg.mag(eph)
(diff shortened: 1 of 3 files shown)


sbpy/photometry/core.py
```diff
@@ -777,7 +876,8 @@ class LinearPhaseFunc(DiskIntegratedModelClass):
 
     _unit = 'mag'
     H = Parameter(description='Absolute magnitude')
-    S = Parameter(description='Linear slope (mag/rad)')
+    S = Parameter(description='Linear slope (mag/deg)')
+    input_units = {'x': u.deg}
 
     @staticmethod
     def evaluate(a, H, S):
```
mommermi: can you add a short docstring here? Any function or method that does not have a leading underscore will show up in the API reference, so people might wonder what this function actually does. Even if it is a 'private' function/method, adding a docstring is nice for future debugging...

sbpy/photometry/core.py
```diff
@@ -848,7 +957,13 @@ def _hgphi(pha, i):
 
     @staticmethod
     def evaluate(pha, hh, gg):
-        return hh-2.5*np.log10((1-gg)*HG._hgphi(pha, 1)+gg*HG._hgphi(pha, 2))
+        func = (1-gg)*HG._hgphi(pha, 1)+gg*HG._hgphi(pha, 2)
```
mommermi: docstring

sbpy/photometry/core.py
```diff
@@ -960,7 +1089,13 @@ def _G2(self):
 
     @staticmethod
     def evaluate(ph, h, g1, g2):
-        return h-2.5*np.log10(g1*HG1G2._phi1(ph)+g2*HG1G2._phi2(ph)+(1-g1-g2)*HG1G2._phi3(ph))
+        func = g1*HG1G2._phi1(ph)+g2*HG1G2._phi2(ph)+(1-g1-g2)*HG1G2._phi3(ph)
```
mommermi: docstring

sbpy/photometry/core.py
```diff
@@ -1046,7 +1191,56 @@ def fit_deriv(ph, h, g):
...
-        ddg = 1.085736205*((phi3-phi1)*p1+(phi3-phi1)*p2)/dom
+        ddg = 1.085736205*((phi3-phi1)*p1+(phi3-phi2)*p2)/dom
+        return [ddh, ddg]
+
+    def _parameter_units_for_data_units(self, inputs_unit, outputs_unit):
+        return OrderedDict([('H', outputs_unit['y']),
+                            ('G12', u.dimensionless_unscaled)])
+
+
+class HG12_Pen16(HG12):
```
mommermi: is this the latest-greatest version of G12? If so, should we make it the default?

sbpy/photometry/core.py
```diff
@@ -551,52 +573,142 @@ def phaseint(self):
         """Phase integral"""
         return self._phase_integral()
 
-    def fit(self, eph):
-        """Fit photometric model to photometric data stored in sbpy.data.Ephem
-        object
+    def fit(self, eph, mag, fitter=None, return_fitter=False, **kwargs):
```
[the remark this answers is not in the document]
mommermi: Ok, I think I got it now: `fit` is used for fitting an existing object to data, whereas `from_data` creates a new object from data. 

Feel free to ignore this comment!

sbpy/photometry/core.py
```diff
@@ -26,6 +27,24 @@
 from ..units import hundred_nm
 
 
+def _process_ephem_input(eph, key=None):
+    if eph is None:
+        return None, None
+    if not isinstance(eph, (Ephem, dict, np.ndarray, Number, u.Quantity)):
+        raise ValueError('`~sbpy.data.Ephem`, `dict`, `numpy.ndarray`, `Number`, or `~astropy.units.Quantity expected, {0} received'.format(type(eph)))
+    if isinstance(eph, dict):
+        eph = Ephem(eph)
```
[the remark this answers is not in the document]
jianyangli: Actually initialization directly with `dict` does work.  But do you advise we should always use `.from_dict` for increased API stability?  If so, I can change the code.

sbpy/photometry/core.py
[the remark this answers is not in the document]
jianyangli: I believe so, because uncertainty is automatically supported by the `weights` keyword of almost all fitter classes.  But I will need to test it and also add this into the docs.

sbpy/photometry/core.py
```diff
@@ -551,52 +573,142 @@ def phaseint(self):
...
+        This is a wrapper that provides a consistent interface that is
+        compatible with the `sbpy` guideline.
 
         Parameters
         ----------
-        eph : `~sbpy.data.Ephem` instance, mandatory
-            photometric data; must contain `phase` (phase angle) and `mag`
-            (apparent magnitude) columns; `mag_err` optional
+        eph : `~sbpy.data.Ephem`, dict_like, or array_like of float
+            If `~sbpy.data.Ephem` or dict_like, must contain a column with a
```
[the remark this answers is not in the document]
jianyangli: Will do.

sbpy/photometry/core.py
```diff
@@ -551,52 +573,142 @@ def phaseint(self):
...
+            name that's compatible with the alternative names for phase angle
+            (see `~sbpy.data.DataClass`).  If array_like, then it is the phase
+            angles of object.  If any distance (heliocentric and geocentric)
+            is provided, then they will be used to correct magnitude to
+            1 au before fitting.  If no unit is provided via type
+            `~astropy.units.Quantity`, then radiance is assumed for phase
```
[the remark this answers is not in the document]
jianyangli: This is because the base functions in the IAU H, G and other systems are defined in radians.  I can set degrees as the default unit for angles and add the degree-to-radians conversion into `.evaluate` and `fit_deriv`, but then the code will run nearly 4x slower when the input angle is dimensionless, and about 1.5x slower for Quantity input.

So what's your suggestion for this trade off?

sbpy/photometry/core.py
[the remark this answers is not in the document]
jianyangli: Correct.  I debated about setting `.fit` as private, but then decided to make it public so users can also play with it.

sbpy/photometry/core.py
jianyangli: Well, I don't fully understand it yet, but it appears to control the default unit of input `x`, based on the docs in `astropy.modeling.Model` and other examples that I could find online.  Will add this to the docstring.

sbpy/photometry/core.py
jianyangli: Do we really need docstrings for all the `.evaluate` and `.fit_deriv` method?  For people who care about defining their own model classes, they would understand what these are; for those who just need to use the model classes we provided here, they never need to directly call these methods.

sbpy/photometry/core.py
jianyangli: I don't know.  I think the difference is that this new version doesn't have discontinuity in the `G12` parameter and its 1st derivative.  But I have no idea which one is "more official" or "better".  Let me ask Antti.

jianyangli:
@mommermi Thanks for your review and suggestions.

I will try to allow `DataClass` to initialize `DiskIntegratedPhaseFunc` classes, but I suspect this will involve some large change to the code.  How about me exploring this more in a separate branch in the next few days, then we can decide whether we include it in v0.1 release?

Regarding field names, since I'm mostly a user of the alternative names,  I'm happy with the current list you have in the coding.  Will let you know if I notice any names on the online databases that are not in your list.

sbpy/photometry/core.py
jianyangli: Well, how do we add docstring to attribute like this?  @mommermi   I tried below code but the docstring doesn't show up in `help(LinearPhaseFunc)`.

```
input_units = {'x': u.deg}
"""Default unit for independent variable"""
```

sbpy/photometry/core.py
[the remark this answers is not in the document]
mkelley: How about making a copy of the `Ephem` object and converting the phase column to radians?

sbpy/photometry/core.py
mkelley: I think via: `input_units.__doc__ = "documentation"`

mkelley: Nevermind: `AttributeError: 'dict' object attribute '__doc__' is read-only`

sbpy/photometry/core.py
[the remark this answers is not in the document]
jianyangli: We could do this, but then the question is, do we want to do this for all methods, including overloading the `.__call__` method?  If we do so, then there will be inconsistency and confusion from attribute `.input_units` because it specifies the default units taken by the static method `.evaluate` as defined by the base class `astropy.modeling.Model`...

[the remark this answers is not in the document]
mkelley: The point of this method is to provide a convenience for the user with `sbpy` classes, right?  Seems like we should be requiring something that is compatible with the `Ephem` class.  That class should require units for these parameters.  What the other methods do can be treated separately, right?

sbpy/photometry/core.py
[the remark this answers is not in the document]
mommermi: Hm. Interesting. I guess it works because `sbpy.data.DataClass` takes `*kwargs` as input parameter and then interprets the dictionary correctly. Yeah, maybe change it to use `.from_dict` - just to be on the safe side... Thanks!

sbpy/photometry/core.py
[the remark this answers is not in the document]
mommermi: Can you make `evaluate` and all `astropy.modeling` aspects work with radians and then do a conversion in the higher level methods that the user is calling? It's not really the cleanest solution, but I think users would really prefer to work with angles instead of radians...

mommermi:
Thanks, @jianyangli! Opening a new branch for testing sounds fine to me. Implementing it shouldn't be too hard. Just keep an internal `self.phys` object and grab the radius (or anything else like albedos, etc.) from that. If you need help, let me know!

sbpy/photometry/core.py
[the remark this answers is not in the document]
jianyangli: @mkelley we are trying to make as many methods as possible to work with `sbpy` classes, so it's not just a few methods that we treat differently.  But if we change the default of all methods, then we might need to touch the core of `astropy.modeling.Model`, which we should definitely avoid.

@mommermi True that I like to work in degrees rather than radians.  But I'm really worried that different default unit for different methods will potentially create serious confusions.  In my opinion, setting default as 'rad' isn't a big issue for users, because all methods (including those `astropy.modeling.Model` methods) can take `Quantity` input, and the `Ephem` object, at least when generated from online sources using, e.g., `.from_horizons` all have phase angles in `Quantity`.  If users are not sure, they can just explicitly supply a unit to phase angle.  On the other hand, leaving default as 'rad' keeps the code clean.  So my preference is still let 'rad' as default.

sbpy/photometry/core.py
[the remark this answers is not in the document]
jianyangli: Sure.

sbpy/photometry/core.py
```diff
@@ -551,52 +573,142 @@ def phaseint(self):
...
+    def _distance_module(self, eph):
+        """Return the correction magnitude or factor for heliocentric distance
+        and observer distance
 
-        not yet implemented
+        Parameters
+        ----------
+        eph : any type
+            If `~sbpy.data.Ephem` or dict_like, then it's used to calculate
+            distanc correction.  If no unit is provided via type
```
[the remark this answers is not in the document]
jianyangli: In fact there is no required fields.  The program will search for relevant fields ('rh' and 'delta' or the equivalent) and calculate the correction if exist.  If a field doesn't exist, then no correction will be included for that field.

I revised the docstring for `eph`.  Please let me know if it's clear now.

sbpy/photometry/core.py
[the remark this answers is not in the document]
jianyangli: Yes we can use `weights` keyword for all the fitter classes currently defined in `astropy.modeling.fitting` submodule.  I will add this note to the docstring.

sbpy/photometry/core.py
jianyangli: Instead of docstring, I added comment lines in the class definition of `DiskIntegratedPhaseFunc`.

jianyangli:
@mommermi I added initialization keyword `data=` to take `sbpy.data.DataClass` to initialize model class.  It's not as complicated as I thought, so I merged it into this PR.  I don't keep a copy of `self.phys` internally to save memory, but just copied the target names into `.meta['targetname']`.  Could you please take a look?  If this implementation fine with you, I'll then update the tests.  Thanks.

sbpy/photometry/core.py
```diff
@@ -26,6 +27,24 @@
 from ..units import hundred_nm
 
 
+def _process_ephem_input(eph, key=None):
```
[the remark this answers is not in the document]
jianyangli: Added.

sbpy/photometry/core.py
```diff
@@ -946,9 +1065,19 @@ class HG1G2(HG12BaseClass):
...
 
-    H = Parameter(description='H parameter')
-    G1 = Parameter(description='G1 parameter')
-    G2 = Parameter(description='G2 parameter')
+    H = Parameter(description='H parameter', default=8)
+    G1 = Parameter(description='G1 parameter', default=0.2)
+    G2 = Parameter(description='G2 parameter', default=0.2)
+
+    @G1.validator
+    def G1(self, value):
```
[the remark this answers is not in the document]
jianyangli: added

sbpy/photometry/core.py
```diff
@@ -946,9 +1065,19 @@ class HG1G2(HG12BaseClass):
...
+        if np.any(value < 0) or np.any(value + self.G2 > 1):
+            warnings.warn('G1, G2 parameter combination might result in a non-monotonic phase function', RuntimeWarning)
+
+    @G2.validator
+    def G2(self, value):
```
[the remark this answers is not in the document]
jianyangli: added

sbpy/photometry/core.py
```diff
@@ -995,8 +1135,13 @@ class HG12(HG12BaseClass):
 
     """
 
-    H = Parameter(description='H parameter')
-    G12 = Parameter(description='G12 parameter')
+    H = Parameter(description='H parameter', default=8)
+    G12 = Parameter(description='G12 parameter', default=0.3)
+
+    @G12.validator
+    def G12(self, value):
```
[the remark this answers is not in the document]
jianyangli: added

sbpy/photometry/core.py
```diff
@@ -507,12 +549,104 @@ class DiskIntegratedModelClass(Fittable1DModel):
     Bond albedo is 0.0184
     >>> print('Phase integral is {0:.3}'.format(phaseint))
     Phase integral is 0.368
+
+    Initialization with subclass of `~sbpy.data.DataClass`:
```
mommermi: This. Is. So. Cool! Thanks!

jianyangli: But then we have a problem here:  Now one model class object could contain multiple phase function models, it will calculate the magnitudes for multiple targets at once.  The question is, how do save the magnitudes of multiple targets into one single `Ephem` object for output?

jianyangli:
I don't understand the Travis-CI failure.  Could someone help?  Thanks.

sbpy/photometry/core.py
mkelley: Following the astropy docs, the `mag` method should allow multiple `Ephem` objects:
> When evaluating a model set, by default the input must be the same length as the number of models, with one input per model.

If only one is provided, then return multiple `Ephem` objects.

Are there use cases where returning a modified input object will be awkward?  Let the model set represent V, R, and I phase functions.  It wouldn't make sense to return three different ephemeris objects.  Let the `Ephem` object contain multiple asteroids.  If the user instantiated the model with it, then wouldn't they expect the `mag` method to apply the appropriate model to the corresponding lines in `eph`?  Also, what if the `Ephem` object already has a magnitude column?

I think a simpler approach is to return `Quantity` object(s) that the user can append as a column on their own.

jianyangli: @mkelley Right, allowing model sets would make things extremely complicated, which we should avoid I believe in order to ensure the robustness of our module.  I do think we want to allow model sets because it would eventually facilitate fitting multiple asteroids/bands all at once.  This is in fact the whole point of model sets:
> Model sets are used primarily for fitting, allowing a large number of models of the same type to be fitted simultaneously (and independently from each other) to some large set of inputs.

So I tend to agree with your suggestion to go back to just returning `Quantity` object(s) and let user decide how to use them.  What do you think, @mommermi ?

mommermi: I follow the reasoning of @mkelley - up to the final part :) I encountered the same problem in `sbpy.thermal` and took the following approach:
* for now, a model class instantiation can only contain a single object; `sbpy.thermal` hence takes a single-line `Phys` object and one `Ephem` object corresponding to the former
* the fitting functions (`.fit` and `.from_data` like in `sbpy.photometry`) return an updated `Phys` object; the `.flux` method returns an updated `Ephem` object

"Updated" means here that information resulting from the methods are added to the input object tables. I prefer this approach as it preserves the order of the data and minimizes confusion. Especially in the case of thermal (and probably also photometric) data, the resulting properties (like flux) are often also only meaningful if the other input parameters are available at the same time. Yes, people could append a `Quantity` object easily to a table, but I think that is unnecessary overhead. Whatever you want to do with the resulting properties (e.g., plotting), you will always need the other properties that are present in the input object. Hence, I think we should always keep everything together and simply append the results. 

Does this make sense?

mkelley: What is your approach for a spectrum?  If I input 500 wavelengths, then the ephemeris will come back with 500 extra columns?

mommermi: I am not adding a lot of columns here. Instead - in the case of `sbpy.thermal` - I simply add two columns (`thermal_lam` for wavelengths and `thermal_flux` for fluxes) and then project the corresponding data on these columns. I wrote a `DataClass.expand` function just for this purpose (https://github.com/mommermi/sbpy/blob/thermal/sbpy/data/core.py#L521). 

In the case of a single spectrum with 500 wavelengths, the table will have a length of 500 and a minimum of 2 columns (if we drop all ephemeris information, which could be stored separately).

mkelley: Ah, so you are dropping the ephemeris data.  Why return an `Ephem` object?

mommermi: I'm not dropping them in the case of `thermal` data, since there's usually not a lot of data points - I guess the same applies to `photometry` data. I'm just saying that we could drop the ephemeris data in the case of a spectrum. Simply because the overhead would be big.

mommermi:
@jianyangli, there are two things that are causing trouble:
* `/home/travis/build/NASA-Planetary-Science/sbpy/docs/sbpy/photometry.rst:148: WARNING: more than one target found for cross-reference 'fit': sbpy.photometry.DiskIntegratedPhaseFunc.fit, sbpy.shape.Lightcurve.fit, sbpy.spectroscopy.Spectrum.fit, sbpy.thermal.ThermalClass.fit`
In this case `.fit` is ambiguous as it exists in several namespaces; try using `~sbpy.photometry.DiskIntegratedPhaseFunc.fit` instead (also, please do the same for `.from_data`)
* `/home/travis/build/NASA-Planetary-Science/sbpy/docs/sbpy/photometry.rst:165: WARNING: Block quote ends without a blank line; unexpected unindent.`
This one is trickier. It might have to do with the fact that you are using backticks in a comment, which might confuse sphinx. Try removing the backticks. If that does not help, there are other things we can try.

sbpy/photometry/core.py
mkelley: There's not a lot of data points, unless you are like me and work with a lot of data points :)  For example, we have ground-based mid-IR spectra of comet 46P, and we want to have the nucleus removed before we model the dust composition.  Or, consider IRTF/SpeX LXD spectroscopy of an NEA.  The thermal emission will be significant throughout the entire wavelength range.  I recommend a keyword argument that controls whether or not a Quantity is returned.  My preference would be to have Quantity the default, because that is the simple case.

I do like the ability to have flux densities and such appended to ephemeris objects.  In fact, I have already been doing something similar in mskpy.  There, all my models returned flat arrays, but higher-level objects could generate lightcurve tables from ephemeris data.  Having that option in sbpy is a must.

This has me thinking about combining models.  We will want to combine thermal and reflected light, accounting for phase effects and reddening, and then use the combined model to fit photometry, a spectrum, or lightcurve.  Given all this subclassing and overloading, it will still work, right?

jianyangli: Thanks guys for the suggestions.  Looks like we have to defer to the future for a full resolution of the issues involved here.  This can be a good topic of discussion when we all get together in June.  For v0.1 release, how about this approach:

1. We instantiate a model class using only the first valid parameter set found in the input `DataClass` object.  We do not deal with multiple objects and model set at this time.
1. We return an `Ephem` object with the high level methods, such as `.mag`, `.ref` in photometry model classes.  At least I don't have to deal with the multiple wavelength cases with one model class like in @mommermi 's `thermal` module :)

Please let me know if this sounds good to you for v0.1 @mkelley @mommermi

jianyangli:
@mommermi I fixed the problems you listed here for `photometry.rst`, but still failed.  Could you take a look again?  Thanks.

sbpy/photometry/core.py
mommermi: Ok, so how about this:
* as @mkelley suggested, functions and methods can return `Quantity` objects if they only calculate a single property. If they calculate several properties, they should be bundled in a `DataClass` table. Even if a function/method only returns a single property, there should always be an option to add the newly derived properties to the corresponding input `DataClass` object (if any)
* as @jianyangli  suggested, let's only support single object models for now.

Can we all agree on this approach? If so, I can update the wiki page.

Concerning combining models: this will be tricky. `sbpy.thermal` is currently supporting phase effects only for the NEATM. I have been thinking a lot about how to setup a model for reflected solar light and albedo calculations using phase integrals. For now, I will all this hard code into my code for testing, but the goal is to use existing or build new models for these cases. That is something we can also aim for in June.

mommermi:
@jianyangli This is the current issue:

> ______________________ [doctest] docs/sbpy/photometry.rst ______________________
> 099 the H-parameter.  The default is the apparent V-magnitude of the Sun,
> 100 M_sun = -26.74 mag.
> 101 
> 102 The model class can also be initialized by a subclass of ``sbpy``'s
> 103 `~sbpy.data.DataClass`, such as `~sbpy.data.Phys`, as long as it contains the
> 104 model parameters:
> 105 
> 106   >>> from sbpy.data import Phys
> 107   >>> phys = Phys.from_sbdb(['Ceres', 'Pallas'])
> 108   >>> m = HG(data=phys)
> Expected:
>     INFO: Model set initialized with 2 objects. [sbpy.photometry.core]
>     INFO: See `.meta['targetname']` for objects included. [sbpy.photometry.core]
> Got:
>     INFO: Model initialized for 1 Ceres. [sbpy.photometry.core]

This is related to your changes in 75b5c13 . In your new implementation, only the first object in the list is considered, but the docstring still expects two objects.

jianyangli:
@mommermi Thanks.  But that was a different one from what I referred earlier.  I fixed them and pushed again.  Could you please take a look at this failure at 476.7?  I can't find any error message in the job log.

sbpy/photometry/core.py
jianyangli: Sounds good to me.  I'll update the code to follow this approach.  Do you have a suggestion for the keyword name to return a `DataClass` object so we can all use the same name across the whole `sbpy` package?

mommermi: how about `append_results`?

mkelley: Yes, let's take this route for now and discuss more in June.

jianyangli:
OK, I believe I've done everything needed for v0.1.  @mommermi @mkelley Please let me know if you are fine with the code now and I'll then (rebased and) merge it to master.

mommermi:
@jianyangli, is this ready to be merged?

jianyangli:
@mommermi I thought you and/or Mike wanted to give it another pass.  Otherwise yes, we can merge it.

mommermi:
Looks good to me! Only one comment: the current version requires `targetname` to be populated if a `DataClass` object is provided; I modified the code slightly to make this optional. Do you want me to push this to your branch?

jianyangli:
Yes please.  I'll take a look and then merge it.  Thanks for checking.

PR #136, opened 2019-04-18
Reflectance conversion rebased

jianyangli:
This PR replace #128


jianyangli:
@mkelley Thanks for the T1 spectrum.  The reflectance values calculated based on the spectrum near 1 µm are consistent with the value that I estimated based on the Hapke photometric model in the visible.  So that's good.

I implemented everything (I think) that you suggested in the new PR, and am satisfied with it now.  Please take a look when you have time.  Thanks.

sbpy/units/core.py
```diff
@@ -133,3 +136,150 @@ def spectral_density_vega(wfb):
...
+    """Reflectance related equivalencies.
+
+    Support conversion from/to reflectance and scattering cross-section
+    to/from total flux and magnitude.
+
+    The default magnitude system is ``VEGAmag``, where the apparent magnitude
+    of Vega is assumed to be 0 in all and any wavelengths and bands.  If other
+    magnitude system is used, then it is implicitly inferred from the solar
+    magnitude passed by keyword 'M_sun'.
+
```
mkelley: It is covered in the `M_sun` documentation, but it may help the user to quote the default values in the summary, and that these are not affected by the sbpy solar spectrum calibration system.  If the user wants something based on the current default solar spectrum, they need to provide `wfb`.

sbpy/units/core.py
```diff
@@ -133,3 +136,150 @@ def spectral_density_vega(wfb):
...
+        f_sun=['W/(m2 um)', 'W/(m2 Hz)'])
+def reflectance(cross_section=None, reflectance=None, wfb=None, M_sun=None,
+        f_sun=None):
+    """Reflectance related equivalencies.
+
+    Support conversion from/to reflectance and scattering cross-section
+    to/from total flux and magnitude.
+
+    The default magnitude system is ``VEGAmag``, where the apparent magnitude
+    of Vega is assumed to be 0 in all and any wavelengths and bands.  If other
```
mkelley: "If [another]"

sbpy/units/core.py
```diff
@@ -133,3 +136,150 @@ def spectral_density_vega(wfb):
...
+    magnitude system is used, then it is implicitly inferred from the solar
+    magnitude passed by keyword 'M_sun'.
```
mkelley: "passed by keyword `M_sun`." i.e., change the quotes.

sbpy/units/core.py
```diff
@@ -133,3 +136,150 @@ def spectral_density_vega(wfb):
...
+
+    Parameters
+    ----------
+    cross_section : `astropy.units.Qauntity`
+        Total scattering cross-section
+    reflectance : `astropy.units.Quantity`
+        Average reflectance
```
mkelley: One of `cross_section` or `reflectance` is required, right?

sbpy/units/core.py
```diff
@@ -133,3 +136,150 @@ def spectral_density_vega(wfb):
...
+        used.
+
+    Returns
+    -------
+    eqv : list
+        List of equivalencies
+
+    Examples
+    --------
+    >>> # Convertion between scattering cross-section and reflectance
```
mkelley: Note these examples use built-in constants for V-band.

sbpy/units/core.py
```diff
@@ -133,3 +136,150 @@ def spectral_density_vega(wfb):
...
+            warn(AstropyWarning('synphot required for Vega-based magnitude'
+                            ' conversions.'))
+            return []
+        sun = Sun.from_default()
+        try:
+            f_sun_lam = sun.observe(wfb, unit='W/(m2 um)')
+            f_sun_nu = sun.observe(wfb, unit='W/(m2 Hz)')
+        except SinglePointSpectrumError:
+            f_sun_lam = sun(wfb, unit='W/(m2 um)')
+            f_sun_nu = sun(wfb, unit='W/(m2 Hz)')
```
mkelley: How about raising a warning here since interpolation of the solar spectrum is being used?

sbpy/units/core.py
```diff
@@ -133,3 +136,150 @@ def spectral_density_vega(wfb):
...
+    >>> print('{0:.2f}'.format(mag2))
+    3.40 mag(VEGA)
+    """
+    if wfb is not None:
+        try:
+            import synphot
+        except ImportError:
+            warn(AstropyWarning('synphot required for Vega-based magnitude'
+                            ' conversions.'))
+            return []
```
mkelley: The try...except is not needed since `sun.from_default` should raise an ImportError (either now or when my next PR is merged).

sbpy/units/tests/test_core.py
```diff
@@ -96,3 +97,83 @@ def test_spectral_density_vega_bp(filename, fluxd, to, tol):
...
+    spec = (t1['spec'] * u.Unit('W/(m2 um sr)') * ifov**2).to('W/(m2 um)',
+        u.dimensionless_angles()) * delta.to('au').value**2 \
+        * rh.to('au').value**2
+    spec_nu = spec.to('W/(m2 Hz)', u.spectral_density(wave))
+    xsec = (ifov * delta).to('km', u.dimensionless_angles())**2
+    ref1 = spec.to('1/sr', reflectance(cross_section=xsec, wfb=wave))
+    ref2 = spec.to('1/sr', reflectance(cross_section=xsec, f_sun=t1['flux_sun']*u.Unit('W/(m2 um)')))
+    ref3 = spec_nu.to('1/sr', reflectance(cross_section=xsec, f_sun=t1['flux_sun_nu']*u.Unit('W/(m2 Hz)')))
+    assert ref1.unit == '1/sr'
+    assert np.isclose(ref1.value, t1['ref']).all()
```
mkelley: FYI, there is also an `np.allclose()` function.

sbpy/units/tests/test_core.py
```diff
@@ -96,3 +97,83 @@ def test_spectral_density_vega_bp(filename, fluxd, to, tol):
...
+    assert ra.unit == u.km
+    assert np.isclose(ra.value, radius)
+
+
+def test_reflectance_spec():
+    """Test conversion from flux spectrum to reflectance spectrum
+
+    Use Tempel 1 spectrum collected by Deep Impact HRI-IR at
+    2005-07-04 05:32 UT as the standard.  Data file
+    'data/hi05070405_9000036-avg-spec.txt'
```
mkelley: Add citation information here?

sbpy/units/tests/data/hi05070405_9000036-avg-spec.txt
```diff
@@ -0,0 +1,527 @@
...
+# Planetary Data System.
+#
+# Mid time ~ 2005-07-04 05:32 UT
+#
+# Averaged over illuminated area (29.81 km**2) defined using a pixel threshold
+#
+# J-Y Li 2019 Apr 18
+#
+# Added columns 'flux_sun', 'flux_sun_nu', ref
+# Solar fluxes are calculated with sbpy.spectroscopy.Sun.from_default().observe()
```
mkelley: Note the actual spectrum used in case the default changes in the future.

sbpy/units/core.py
mkelley: Shouldn't the magnitude system be explicitly specified in the `.to` method?

sbpy/units/core.py
mkelley: Instead, can the equivalencies be expressed in per unit area and per unit reflectance so that the user does not need to add them to the `reflectance` call?  I guess in that case `rh` and `delta` are needed, right?

sbpy/units/core.py
jianyangli: Done.

sbpy/units/core.py
jianyangli: But there is no way for equivalency function to know what's passed to `.to`.

sbpy/units/core.py
jianyangli: Well, we can express the scattering cross-section in terms of solid angle, and that's what you actually mean by "per unit area", right?  I think this is kind of overdoing (or maybe underdoing) and requires one extra step for users to understand this equivalency, and therefore defeats the purpose of providing convenience to users.

I debated about adding `rh` and `delta` into the equivalence call, and decided not, because then we'll also need to add a phase function model to the parameter list.  That's going to be too much of integration and will make this equivalency too complicated.  I think we want to keep it clean and simple.

sbpy/units/core.py
jianyangli: Added.

sbpy/units/core.py
jianyangli: I don't understand the part starting from "..., and that these are not affected ..." through the end.

sbpy/units/core.py
jianyangli: Good idea.  Is there a pre-defined warning in sbpy.spectroscopy that I can use?

sbpy/units/tests/test_core.py
jianyangli: Sure.  What's the best citation to this spectrum because you sent it to me :)

sbpy/units/tests/data/hi05070405_9000036-avg-spec.txt
jianyangli: Good point.  Reference added.

sbpy/units/tests/test_core.py
jianyangli: Revised for `np.allclose()`.  Thanks for pointing it out to me.

sbpy/units/core.py
mkelley: Sounds good.

sbpy/units/core.py
jianyangli: Got it.  Thanks.

sbpy/units/core.py
mkelley: No, I haven't defined any.

sbpy/units/tests/test_core.py
mkelley: It is in the file header.

jianyangli: Yes, sorry for missing that.

sbpy/units/core.py
jianyangli: I'll just use `AstropyWarning` then.

sbpy/units/core.py
mkelley: I don't think there is any default magnitude system, but just a default bandpass and solar model.  I can use this to convert a magnitude in the ABmag system without specifying M_sun.  This is because f_sun_nu is defined.  I think the note should instead be: "The default bandpass is Johnson V and solar model is XXX.  If the user does not specify a wavelength, frequency, or bandpass via `wfb`, then VEGAmag and spectral flux density equivalencies will be provided for the V-band."  If you want to say more: "Users wanting conversions with magnitude systems incompatible with these units will need to define `M_sun` in that system or add equivalencies to convert from the magnitude system to physical units or Vega-based magnitudes."

mkelley:
This is OK for the moment, but I think we should avoid hard coding any photometric values in any function.  I'm trying to figure out the right approach, and we can discuss at another time.

sbpy/units/core.py
jianyangli: Now it makes sense to me.  Thanks for the suggestion.

jianyangli:
OK I need someone to help again with the Travis-C tests.  It complains that the test file `units/tests/data/hi05070405_9000036-avg-spec.txt` not found, but all related tests passed on my local machine.  Someone helps me to fix it?  Thanks.

PR #142, opened 2019-04-30
Fix value of spectral flux density in remote test

migueldvb:
This change fixes a bug in a remote test. Hopefully it can go into v0.1 if there is still time.


mommermi:
... plenty of time ... thanks!

PR #164, opened 2019-06-26
Overhaul of sbpy.activity.gas into its own directory. Changed mol_data to Phys.from_jplspec() 

gianninapr:
Overhaul of production rate functions into sbpy.activity and the adoption of mol_data function from sbpy.spectroscopy into a Phys object. Included in another future PR will be the change of accepted inputs from the production rate functions into phys objects, allowing user more flexibility if they do not want to use jplspec.


sbpy/activity/gas/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .. core import *
```
mkelley: This should not be needed.

sbpy/activity/gas/core.py
```diff
@@ -29,6 +29,7 @@
 
     'Haser',
     'Vectorial',
+    'GasComa'
```
mkelley: I left this out of __all__ because it is an abstract class that a user probably doesn't need.  What do you think?

sbpy/data/phys.py
```diff
@@ -12,8 +12,11 @@
 from collections import OrderedDict
 
 from numpy import ndarray, array, isnan, nan
+from scipy import interpolate
```
mkelley: scipy is an optional package.  This should be wrapped to catch an import error, or the import moved to the method that needs it.

sbpy/data/phys.py
```diff
@@ -153,6 +156,99 @@ def from_sbdb(cls, targetids, references=False, notes=False):
...
+        quantities = [t_freq, temp, lgint, part300, partition, gu, energy_J, elo_J, df]
+        names = ('Transition frequency',
+                 'Temperature',
+                 'Integrated line intensity at 300 K',
+                 'Partition function at 300 K',
+                 'Partition function at designated temperature',
+                 'Upper state degeneracy',
+                 'Upper level energy in Joules',
+                 'Lower level energy in Joules',
+                 'Degrees of freedom')
```
mkelley: These names and some short-hand versions should be added to the field name list in `data/__init__.py`

gianninapr: Done!

sbpy/data/phys.py
gianninapr: Taken care of within the method that needs it

gianninapr:
@mkelley I think it's best if this is merged when ready. I can just open another PR that covers the having the molecular data phys object as the input type for the production rate functions. The reason being that it's going to take me a bit to figure out the implementation of that since it's conflicting with how I wrote the code originally, and I think that this PR should be implemented as soon as possible in order to avoid conflicts since it moved a lot of code to an entirely different module

sbpy/data/phys.py
```diff
@@ -9,18 +9,29 @@
...
+from astroquery.jplspec import JPLSpec
 
 from .core import DataClass
 from .. import bib
+from ..exceptions import SbpyWarning
 
 __all__ = ['Phys']
 
 
+class JPLSpecQueryFailed(SbpyWarning):
```
mkelley: I looked at the code raising this exception, and they are fatal errors, i.e., the code cannot complete as requested.  Instead of `SbpyWarning` I recommend using `SbpyException`.

sbpy/data/phys.py
```diff
@@ -153,6 +164,136 @@ def from_sbdb(cls, targetids, references=False, notes=False):
...
+        mol = cat[cat['TAG'] == mol_tag]
+
+        temp_list = cat.meta['Temperature (K)'] * u.K
+
+        part = list(mol['QLOG1', 'QLOG2', 'QLOG3', 'QLOG4', 'QLOG5', 'QLOG6',
+                        'QLOG7'][0])
+
+        temp = temp_estimate
+
+        f = interpolate.interp1d(temp_list, part, 'linear')
```
mkelley: Linear interpolation can also be done with `numpy.interp`.  If you use numpy instead, scipy wouldn't be needed.

mkelley:
After those two changes, I think we can merge!

sbpy/data/phys.py
gianninapr: You're correct, I only recently just learned numpy had a function for linear interpolation, and I can definitely apply it here.

gianninapr:
@mkelley took care of it, also I updated all the documentation!

mkelley:
:+1:

PR #168, opened 2019-06-26
Reorganized docs.

mkelley:


mkelley:
Re-organized documentation to simplify the front page, making it easier for humans to find information.  Most details moved to separate pages.

PR #172, opened 2019-06-29
WIP: photometric calibration

mkelley:
Move sun and vega to new calib; new solar_fluxd, vega_fluxd

Rename old context managers from default_sun to solar_spectrum and
default_vega to vega_spectrum.

solar_fluxd and vega_fluxd allow sbpy calibration without bandpasses
and spectra (or synphot).


mkelley:
I think I am close to getting the new calibration system implemented.

Note that observing through filters named by a string will no longer download the filter in question.  That functionality is covered by synphot, so it isn't an onerous change.  Instead, strings will always refer to the sbpy internal calibration.  For example:

``` python
sun = Sun.from_default()

# old method
sun.observe('johnson_v')  # would download johnson v from STScI and observe through it

# new method
V = synphot.SpectralElement.from_filter('johnson_v')
sun.observe(V)

# why?  because strings now work via:
with solar_fluxd({'V': -26.77 * VEGAmag}):
    sun.observe('V')
```

mkelley:
@jianyangli I've re-written reflectance to handle the new approach.  In doing so, I removed your built-in default values for the Sun, and make wavelength/frequency/bandpass a required input.  Is this OK?  I think this is OK, because you (the user) should be able to load your own defaults at run time:

```python
from sbpy.calib import solar_fluxd
solar_flux.set(my_favorite_values)
...
```
Maybe at some point we can make an sbpy calibration file that could load such values automatically.

However, while fixing the tests, I've hit a snag on `test_reflectance_spec()`: the reflectances at the ends of the spectrum are too large:
```python
(Pdb) ref1.value / t1['ref'].data
array([1.14234968, 1.00034833, 1.00036675, 1.00013488, 1.00002218,
       0.99979961, 0.99974319, 1.00114149, 0.99979567, 0.99937592,
...
       0.99999217, 0.99998612, 0.99998302, 0.99999199, 0.99999026,
       0.99997852, 0.99998371, 0.99998225, 0.99999122, 0.99999046,
       1.14119862])
```

`Sun.observe(wavelengths)` is to blame.  I will try to figure out what is happening here.

jianyangli:
@mkelley Yes I think this is conceptually clear.  After all, all observations are associated with a bandpass or wavelength.  Thanks for implementing this.

In some random test I did, I think there is a bug.  Try below code.  The second line from the last works fine, but the last line throws an error: `` FilterLookupError: Filter for "V" is not present in `vega_fluxd` ``.  But `vega_fluxd` isn't necessary here.  This bug is causing an error for a randomly defined solar_fluxd and bandpass string when used for `reflectance` equivalency.

```
from sbpy.calib import solar_fluxd, Sun
from sbpy.units import VEGAmag, VEGA

solar_fluxd.set({'V': -26.77 * VEGAmag})
sun = Sun.from_default()

# works fine:
sun.observe('V').to(VEGA)

# causes an error
sun.observe('V', unit=VEGA)
```

mkelley:
Good catch.  The code now avoids using `Vega` if the unit conversion does not require it.

mkelley:
I fixed the problem with the reflectance spectra.  The test takes the solar spectrum from the Tempel 1 file and uses it as the solar spectrum for calibration.  The results were good to 0.18% (1σ), but there were a few outliers at the 2% level.  The test is a special case when the native wavelengths of the solar spectrum are the same as the requested wavelengths for an observation.  The error has something to do with how synphot computes the bin edges; the total flux seems to be conserved, but a few neighboring points have high deviations.

I can either (1) add an option to force interpolation in all cases: ``Sun.from_array(t1['wave'], t1['flux_sun_nu'], interpolation=True)``; or (2) when the user requests it: ``sun.observe(wave, interpolation=True)``; or (3) catch the special case when the requested wavelengths are the same as the internal solar spectrum wavelengths and return the spectrum unbinned: ``sun.observe(wave)``.

I don't really like (2), because I have tried to keep a distinction between observations through a spectrometer (rebinning) and interpolation of the spectrum: ``sun.observe(wave)`` vs. ``sun(wave)``.  I currently have both (1) and (3) implemented, but I think (3) is my favorite.

mkelley:
I looked into what synphot is doing for these observations, and the assumption synphot makes is that the spectrum is a continuous function.  However, the solar spectrum is binned.  When this binning is at high spectral resolution, and the user requests a comparatively low spectral resolution, the assumptions within synphot are OK.  But, as pointed out above, when the resolutions are comparable, there can be problems.  Ultimately, I will go with option (2): add an ``interpolate`` option to ``observe``.

To support this request, functions that call ``observe`` need to pass keyword arguments to the ``observe`` call.  I have modified the ``reflectance`` equivalency accordingly.

This problem and the choice between rebinning and interpolation is now explained in the documentation.

mkelley:
Closed in favor of #186 .

PR #174, opened 2019-06-30
Mark tests remote

mkelley:
On the plane yesterday, I found a lot of code fetching remote data, but wasn't marked remote.  This PR attempts to catch all of them so that all tests will pass when run without a network connection.


mkelley:
I've turned on a flag in `setup.cfg` that will prevent remote data access during tests, unless specifically requested.

PR #175, opened 2019-07-03
Changed Partition fn interpolation to log-log space

gianninapr:


mommermi:
remote tests broke due to MPC server being down - merge anyway

PR #189, opened 2019-07-12
WIP: Implementation of Pyradex to obtain column densities from flux data

gianninapr:
Calculate production rates in a Non-LTE fashion by obtaining column densities from RADEX nonLTE iterative fortran code. Remote tests will need to be monkey patched in the future.

### DIFF
--- a/sbpy/activity/gas/productionrate.py
+++ b/sbpy/activity/gas/productionrate.py
@@ -20,8 +20,6 @@ from astroquery.lamda import Lamda
 from ...bib import register
 from ...data import Phys
 
-conf.horizons_server = 'https://ssd.jpl.nasa.gov/horizons_batch.cgi'
-
 __all__ = ['LTE', 'NonLTE', 'einstein_coeff',
            'intensity_conversion', 'beta_factor', 'total_number',
            'from_Haser']
--- a/sbpy/activity/gas/productionrate.py
+++ b/sbpy/activity/gas/productionrate.py
@@ -6,11 +6,13 @@ created on June 26, 2019
 """
 
 import numpy as np
+import tempfile
 import astropy.constants as con
 import astropy.units as u
 from astropy.time import Time
 from astroquery.jplhorizons import Horizons, conf
 from astroquery.jplspec import JPLSpec
+from astroquery.lamda import Lamda
 from ...bib import register
 from ...data import Phys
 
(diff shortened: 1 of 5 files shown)


sbpy/activity/gas/productionrate.py
```diff
@@ -6,11 +6,13 @@
 """
 
 import numpy as np
+import tempfile
```
migueldvb: `tempfile` could be imported first because it is in the standard library

sbpy/activity/gas/productionrate.py
```diff
@@ -618,3 +620,175 @@ class NonLTE():
...
+
+    def from_pyradex(self, integrated_flux, mol_data, line_width=1.0 * u.km / u.s,
+                     escapeProbGeom='lvg', iter=100):
+        """
+        Calculate production rate from the Non-LTE iterative code pyradex
+        Presently, only the LAMDA catalog is supported by this function.
+        In the future a function will be provided by sbpy to build your own
+        molecular data file from JPLSpec for use in this function.
+        Collider is assumed to be H2O since we are only considering comet
+        prodcution rates.
```
migueldvb: production rates

sbpy/activity/gas/productionrate.py
```diff
@@ -618,3 +620,175 @@ class NonLTE():
...
+            name = mol_data['mol_tag'][0]
+            name = name.lower()
+
+        # try various common instances of molecule names and check them against LAMDA before complaining
+        try:
+            Lamda.molecule_dict[name]
+        except KeyError:
+            try_name = "{}@xpol".format(name)
+            try:
+                Lamda.molecule_dict[name]
```
migueldvb: This statement should be `Lamda.molecule_dict[try_name]`

sbpy/activity/gas/productionrate.py
```diff
@@ -618,3 +620,175 @@ class NonLTE():
...
+        except ImportError:
+            raise ImportError('Pyradex not installed. Please see \
+            https://github.com/keflavich/pyradex/blob/master/INSTALL.rst')
+
+        if not isinstance(mol_data, Phys):
+            raise ValueError('mol_data must be a `sbpy.data.phys` instance.')
+
+        register('Production Rates', {'Radex': '2007A&A...468..627V'})
+
+        # convert mol_tag JPLSpec identifier to verbose name if neeeded
```
migueldvb: `if needed`

gianninapr:
@migueldvb  Will take care of these recommendations on Monday!

gianninapr:
@migueldvb if pyradex is installed correctly in your machine, then the following code should work: 
Screen Shot 2019-07-29 at 11 38 14 AM

That's how it looks when I run it in my local machine with pyradex installed and using the code implemented in this PR. The code should take a bit to run (around 30s - a minute) due to the number of iterations in the example (500), you can lower that number. 

Let me know what error you're getting that's not letting you run it. I'm going to be working on adding detailed documentation that details the common errors with pyradex installation too. 

I'll commit soon with the changes you gave me and then put in a bigger commit that moves Haser out of LTE() class since it can also be used with NonLTE(), as well as moves the other column density function from Bockelee Morvan to LTE() class, just to make sure we keep consistency with the physical meaning of the models we're using.

migueldvb:
Thanks, it works perfectly. I could not make it to work using `myRadex` solver but it runs well with standard `radex`.  It could also have been an issue with my installation. Actually this is not a problem in general because I think that not many people will use the `myRadex` version, so we could simply add a warning about this in the documentation.

sbpy/activity/gas/productionrate.py
```diff
@@ -618,3 +620,175 @@ class NonLTE():
...
+            Closest Integrated Flux:[1.2352747] K km / s # doctest: +SKIP
+            Given Integrated Flux: 1.234 K km / s # doctest: +SKIP
+        >>> print(cdensity) # doctest: +SKIP
+            [1.05143086e+14] 1 / cm4
+
+        References
+        ----------
+        Haser 1957, Bulletin de la Societe Royale des Sciences de Liege
+        43, 740.
+
```
migueldvb: You could include a reference here to the radex paper: `2007A&A...468..627V`

gianninapr:
@migueldvb I'll keep that in mind when looking at the documentation. 

Is there another paper aside from the Europa one we could test against? I'm testing against the paper we used for LTE Haser and the errors have actually gone down for the tests of the implementation of pyradex. There's an issue with testing the Europa paper and it's that they change a lot of properties of the RADEX input that won't be as necessary to change in our case (like trying different ranges of collider densities). I wanted to make the implementation of pyradex as flexible and straightforward as possible for the user while maintaining a simple workflow within sbpy and dedicating it to the calculation of column density for the use in production rates. 

In the future I also want to work with making our own molecular data files based on JPLSpec results so that the user can use those for pyradex if they want instead of using LAMDA. But for now, I think this gives enough functionality for the user. 

The remote tests are running well, feel free to take a look and comment on how it looks. 

I'll be adding new documentation tomorrow

gianninapr:
@migueldvb  added documentation and default collider density scaled for H2O/H2 mass ratio (since H2O will be our main collider). Will try to write a test for the Enceladus paper, errors against the CO paper literature values are looking much better with this implementation! 

Do you think this could be part of version 0.2?

sbpy/activity/gas/productionrate.py
```diff
@@ -495,126 +619,212 @@ def from_Drahus(self, integrated_flux, mol_data, ephemobj, vgas=1 * u.km/u.s,
...
+class NonLTE():
+    """
+    Class method for non LTE production rate models
+    Not Yet implemented
+
+    """
+
+    def from_pyradex(self, integrated_flux, mol_data, line_width=1.0 * u.km / u.s,
+                     escapeProbGeom='lvg', iter=100,
+                     collider_density={'H2': 900*8.934}):
```
migueldvb: Here the "standard" scaling relation for collisional rates is based on the reduced mass of the collision system. For deriving H2O-H2O collision rates from H2-H2O coefficients we would apply the square root of the ratio of reduced masses `(mu_H2O/mu_H2)**0.5 = ((18**2/18*2)/((18*2)/(18+2)))**0.5 = 2.2`. This factor reduces the H2-H2O collision rates (dividing rather than multiplying) because of the slower collisions. See the discussion in [this paper](https://arxiv.org/abs/1406.5750) for the case of He-H2 collision rates. The authors conclude that this scaling is not very reliable but probably is our best option.

gianninapr: Got it, thanks!

sbpy/activity/gas/tests/test_prodrate.py
```diff
@@ -15,3 +27,138 @@ def __init__(self, filename):
...
+'''
+
+
+class TestProductionRate:
+
+    def test_simple_prodrate(self):
+        """Test a set of dummy values."""
+
+        temp_estimate = 47. * u.K
+        target = '103P'
```
migueldvb: It seems that the target name is not used in this test function

sbpy/activity/gas/tests/test_prodrate.py
```diff
@@ -15,3 +27,138 @@ def __init__(self, filename):
...
+
+    def test_Haser_prodrate(self):
+        """Test a set of dummy values."""
+
+        temp_estimate = 47. * u.K
+        target = '103P'
+        vgas = 0.8 * u.km / u.s
+        aper = 30 * u.m
+        b = 1.13
+        transition_freq = (265.886434 * u.GHz).to('MHz')
```
migueldvb: `target` and `transition_freq` are not required

sbpy/activity/gas/tests/test_prodrate.py
```diff
@@ -1,3 +1,15 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+import pytest
+import numpy as np
+import astropy.units as u
+from astropy.time import Time
+from .. import Haser, photo_timescale
+from ....data import Ephem, Phys
+from .. import (LTE, NonLTE, einstein_coeff, intensity_conversion, beta_factor,
```
migueldvb: `NonLTE` will be an only remote test (not needed here)?

sbpy/activity/gas/tests/test_prodrate.py
```diff
@@ -1,3 +1,15 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+import pytest
```
migueldvb: pytest import is needed if you are using the context manager or decorator but perhaps not needed here

sbpy/activity/gas/tests/test_prodrate.py
gianninapr: You're correct, thanks!

gianninapr:
@migueldvb @mommermi  this should be good to merge for v0.2! In the future it'd be cool to have a function to write molecular data files for RADEX from jplspec data

sbpy/activity/gas/tests/test_prodrate_remote.py
```diff
@@ -204,3 +199,161 @@ def test_Haser_prodrate():
...
+
+@remote_data
+def test_einsteincoeff_case():
+    # test untested case for einstein coefficient
+
+    temp_estimate = 47. * u.K
+    vgas = 0.8 * u.km / u.s
+    mol_tag = 27001
+    transition_freq = (265.886434 * u.GHz).to('MHz')
+    q_found = []
```
migueldvb: `q_found` line can be deleted

sbpy/activity/gas/tests/test_prodrate_remote.py
```diff
@@ -204,3 +199,161 @@ def test_Haser_prodrate():
...
+
+@remote_data
+def test_intensity_conversion():
+    # test untested case for intensity conversion function
+
+    temp_estimate = 47. * u.K
+    vgas = 0.8 * u.km / u.s
+    mol_tag = 27001
+    transition_freq = (265.886434 * u.GHz).to('MHz')
+    q_found = []
```
migueldvb: same here

sbpy/activity/gas/tests/test_prodrate.py
```diff
@@ -1,3 +1,14 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+import numpy as np
+import astropy.units as u
+from astropy.time import Time
```
migueldvb: `Time` is not needed now

sbpy/activity/gas/tests/test_prodrate_remote.py
```diff
@@ -10,8 +10,8 @@
 from astroquery.jplspec import JPLSpec
```
migueldvb: Could you add a header to identify the license?

 `# Licensed under a 3-clause BSD style license - see LICENSE.rst`

sbpy/activity/gas/productionrate.py
```diff
@@ -1,24 +1,30 @@
...
 import astropy.constants as con
 import astropy.units as u
 from astropy.time import Time
 from astroquery.jplhorizons import Horizons, conf
 from astroquery.jplspec import JPLSpec
+from astroquery.lamda import Lamda
 from ...bib import register
 from ...data import Phys
 
 conf.horizons_server = 'https://ssd.jpl.nasa.gov/horizons_batch.cgi'
```
migueldvb: This conf line seems to be unnecessary. horizons_server should be fixed in recent releases of astroquery and we are not using `astroquery.jplhorizons` directly here.

mommermi:
Thanks, @gianninapr, for dealing with this beast :)

Issue #180, opened 2019-07-05
Allow bib @cite decorator for class?


jianyangli:
Can we allow @cite decorator to work on a class definition, so we don't have to overload `__init__` in subclass with just a single line call to `super().__init__`.  Something like below:

```
class A():
    pass

@cite({'ref': 'bib code 1'})
class B(A):
    pass

@cite({'ref': 'bib code 2'})
class C(A):
    pass
```

mkelley:
Did you try this before?  At least with PR #190, it just works for me.

jianyangli:
Indeed, it does.  I did try it before and it didn't work.  Can't remember details, though.  Anyway, I guess we can close this issue.

mkelley:
I've gotten this far, but notice how the signature of `__init__` is still based on the true class name:
```python
from sbpy import bib


def cite_class(citations):
    def class_decorator(cls):
        class ClassWrapper(cls):
            @bib.cite(citations)
            def __init__(self, *args, **kwargs):
                return super().__init__(*args, **kwargs)

            __module__ = cls.__module__
            __name__ = cls.__name__
            __qualname__ = cls.__qualname__
            __doc__ = cls.__doc__

        return ClassWrapper
    return class_decorator


@cite_class({'ref': 'paper 1'})
class A():
    pass


class B(A):
    pass


print(issubclass(A, A))
print(issubclass(B, A))

print(A)
print(B)

bib.track()
print('\nbib tracking on')
a = A()
b = B()

print(bib.show())
bib.stop()
bib.reset()
```

```
notwo /tmp $ python3 test.py
True
True
<class '__main__.A'>
<class '__main__.B'>

bib tracking on
__main__.cite_class.<locals>.class_decorator.<locals>.ClassWrapper.__init__:
  ref:
    paper 1
```

mkelley:
Ooops, pasted in the wrong window.  Moving to #195 .

Issue #179, opened 2019-07-05
bib report missing class name


jianyangli:
When add a bib entry to a class method, the class name is missing in the bib report.  For example, execute below code piece:

```
from sbpy import bib
from sbpy.bib import cite

class A():
    @cite({'ref': '1989aste.conf..524B'})
    def test(self):
        pass

bib.track()
a = A()
a.test()
print(bib.to_text())
```

It will print:

```
__main__.test:
  ref:
      Bowell, Hapke, Domingue et al. 1989, Asteroids II, 524
```

The class name `A` is missing in the report `__main__.test:`.  Can we add class name?

jianyangli:
Just discovered that a related problem is that the report will merge all entries with the same function name, despite whether they are from the same class or different classes:

```
from sbpy import bib
from sbpy.bib import cite

class A():
    @cite({'ref': '1989aste.conf..524B'})
    def test(self):
        pass

class B():
    @cite({'ref': '2010Icar..209..542M'})
    def test(self):
        pass

class C():
    @cite({'ref': '2016P&SS..123..117P'})
    def test(self):
        pass

bib.track()
a = A()
a.test()
b = B()
b.test()
c = C()
c.test()

print(bib.to_text())
```

will produce output:

```
__main__.test:
  ref:
      Bowell, Hapke, Domingue et al. 1989, Asteroids II, 524
      Penttilä, Shevchenko, Wilkman & Muinonen 2016, Planetary and Space Science, Vol 123, 117
      Muinonen, Belskaya, Cellino et al. 2010, Icarus, Vol 209, 2, 542
```

mkelley:
Both cases were fixed with PR #190 .

Issue #176, opened 2019-07-03
bib clean-up 


mommermi:
Currently, there is a lot of garbage in the `bib` results:

> INFO: Thank you for using sbpy.  Your session results were based on:
> sbpy.bib.tests.test_bib.force:
>   method:
>       Newton 1687, Philosophiae Naturalis Principia Mathematica. Auctore Js. Newton
> test1:
>   track_this:
>       bibcode1
>   software:
>       bibcode2
>   track_this_too:
>       bibcode
> test2:
>   software:
>       bibcode
> test3:
>   track_this:
>       bibcode
>   software:
>       bibcode
> spectroscopy:
>   fdsa:
>       asdf
>   Dust-free template spectrum of Bohlin 2014:
>       Bohlin 2014, The Astronomical Journal, Vol 147, 6, 127
>   E490-00a (2014) reference solar spectrum (Table 3):
>       doi:10.1520/E0490
> sbpy.data.Ephem.from_horizons:
>   data service:
>       Giorgini, Yeomans, Chamberlin et al. 1996, AAS/Division for Planetary Sciences Meeting Abstracts #28, 25.04
> sbpy.data.Phys.from_sbdb:
>   data service url:
>       https://ssd.jpl.nasa.gov/sbdb.cgi
> activity.dust.Afrho:
>   model:
>       Ahearn, Schleicher, Millis et al. 1984, The Astronomical Journal, Vol 89, 579
> activity.dust.Efrho:
>   model:
>       Kelley, Fernández, Licandro et al. 2013, Icarus, Vol 225, 1, 475
> activity.dust.phase_HalleyMarcus:
>   Halley-Marcus phase function:
>       Schleicher & Bair 2011, The Astronomical Journal, Vol 141, 6, 177
>   Halley phase function:
>       Schleicher, Millis & Birch 1998, Icarus, Vol 132, 2, 397
>   Marcus phase function:
>       Marcus 2007, International Comet Quarterly, Vol 29, 39
>  [sbpy.bib.core]

Everybody, please check your submodules.

mkelley:
Looks like at least one bib test is not not leaving tracking disabled.  I'll take care of it.

mkelley:
I think this is fixed.  I haven't seen any extra lines after the tests.

PR #190, opened 2019-07-15
`@cite`: work with classes, better name registration

mkelley:
Using fully qualified names for `@cite` and eliminating cruft after bib tests.

Addresses #180, #179, #176.


sbpy/bib/tests/test_bib.py
```diff
@@ -144,9 +150,50 @@ def force(mass, accelleration):
...
+    reset()
+
+
+def test_cite_class_method():
+    reset()
+
+    class Physics:
+        @cite({'method': '1687pnpm.book.....N'})
+        def force(self, mass, accelleration):
+            return mass * accelleration
```
migueldvb: `acceleration`

sbpy/bib/tests/test_bib.py
```diff
@@ -144,9 +150,50 @@ def force(mass, accelleration):
...
+    reset()
+
+
+def test_cite_class():
+    reset()
+
+    @cite({'method': '1687pnpm.book.....N'})
+    class Force:
+        def __call__(self, mass, accelleration):
+            return mass * accelleration
```
migueldvb: acceleration

sbpy/bib/tests/test_bib.py
```diff
@@ -144,9 +150,50 @@ def force(mass, accelleration):
...
+    stop()
+    reset()
+
+
+def test_cite_class_method():
+    reset()
+
+    class Physics:
+        @cite({'method': '1687pnpm.book.....N'})
+        def force(self, mass, accelleration):
```
migueldvb: `force` could be a static method but not important because it is only used for testing.

sbpy/bib/tests/test_bib.py
mkelley: Thanks.

sbpy/bib/tests/test_bib.py
mkelley: Got it.

PR #218, opened 2019-08-08
Allow unit check for decorator quantity_to_dataclass

jianyangli:
This PR adds an option for `quantity_to_dataclass` decorator to check the unit of input quantity, similar to what `astropy.units.quantity_input` does:

```
import astropy.units as u
import sbpy.data as sbd
@sbd.quantity_to_dataclass(eph=(sbd.Ephem, 'rh', 'length'))
def temperature(eph):
    return 278 * u.K / (eph['rh'] / u.au)**0.5
print(temperature(1 * u.s))
```

Raise exception:

```
UnitsError: Argument 'eph' to function 'temperature' must be in units convertible to 'm'.
```

This is needed in order to enforce quantity input for our functions.  I tried to stack the `astropy.units.quantity_input` decorator with our `quantity_to_dataclass` decorator but it doesn't work when the input is already a `DataClass` object.

This implementation uses some internal functions of `astropy.units.decorators`.  If no objections, I'll merge it in a day.

### DIFF
--- a/sbpy/data/decorators.py
+++ b/sbpy/data/decorators.py
@@ -93,7 +93,8 @@ def quantity_to_dataclass(**kwargs):
                     from astropy.units.decorators import (_get_allowed_units,
                          _validate_arg_value)
                     units = decorator_kwargs[param.name][2]
-                    if not hasattr(units, '__iter__'):
+                    if isinstance(units, str) or (not hasattr(units,
+                            '__iter__')):
                         units = [units]
                     units = _get_allowed_units(units)
                 else:
--- a/sbpy/data/decorators.py
+++ b/sbpy/data/decorators.py
@@ -110,7 +110,7 @@ def quantity_to_dataclass(**kwargs):
                         " object".format(param.name, wrapped_function.__name__,
                         field, dataclass))
                 units = [x['dimension'] for x in [Conf.fieldnames_info[i] for
-                    i,x in enumerate(field_idx) if x]]
+                    i, x in enumerate(field_idx) if x]]
                 if isinstance(units, str) or (not hasattr(units, '__iter__')):
                     units = [units]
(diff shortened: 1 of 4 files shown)


mommermi:
@jianyangli This looks good to me! It is not clear to me how you check the dimension against `Quantity`, but I was wondering if we could use the information from [here](https://github.com/NASA-Planetary-Science/sbpy/blob/788d5e1b5392a92943cb17ae7feb755ab51c49e8/sbpy/data/__init__.py#L14) instead of having to provide our own dimensions in the decorator?

jianyangli:
@mommermi Good point!  Also this can be used to check whether the specified field is legal in `sbpy`.  I'll do it tomorrow (already past midnight in China now).

mommermi:
Thanks! If we need to change labels in the field name list, we can do that easily. Also, if we have to delay release that's fine, too.

mkelley:
Instead, I recommend considering replacing the 'length', 'angular velocity', etc. descriptions with base units (meters, radian per second, etc) that can be directly tested via `unit.is_equivalent`.

I also recommending the testing take place in DataClass, rather than the decorator.  Then all input units can be checked.  This can be a v0.5 milestone.

jianyangli:
@mkelley I totally agree.  We should standardize the dimension column in a way that simplifies the processing in code.  For example, 'length' can be recognized by `astropy.units` as a valid physical type of units, but `velocity` cannot.  For this reason, I had many lines in `quantity_to_dataclass` to process the strings that are not recognizable by `astropy.units`.

And I also agree that we do the test in DataClass, so the test will include all the lines in the field name list table.  v0.5 sounds like a good milestone to do all of these.

I'll file an issue to document these suggestions.

jianyangli:
Issue #221

PR #224, opened 2019-08-12
Monkeypatching pyradex functionality for testing

migueldvb:
Mocking `NonLTE.from_pyradex` to try to improve the code coverage.


mommermi:
@migueldvb I fixed the test, but I find it weird that `nonLTE.from_pyradex` would return a set instead of a `Quantity`. Is that a consequence of your monkeypatching?

migueldvb:
`nonLTE.from_pyradex` returns an Quantity (column density) but the mocked version is just a dictionary with a value to make it simple. Let me make it return a Quantity as well.

mommermi:
Perfect, thanks! Is this ready for merging?

migueldvb:
Yes, it should be ready now.

PR #226, opened 2019-08-14
Mark gas doctests with remote_data flag

migueldvb:
Trying if the doctests that use remote data will pass...


migueldvb:
All the tests are passing so this could go as well into `v0.2` if still possible.

mommermi:
yes it can... Thanks!

PR #227, opened 2019-08-23
Some edits in activity module

migueldvb:
Adding a reference and a few additional updates.


mkelley:
@migueldvb I don't understand the codecov result.  What is /patch vs /project?

migueldvb:
I suspect that the failed codecov test measures only the amount of tested code
in the commited code, and the project status gives the percent variation in
coverage in the whole project.  Probably the second one is the one we really want.

mommermi:
Unfortunately, if your committed code has a slightly lower coverage than the original code, the "patch"  codecov check will always fail. I wish this "feature" was not there and we only had the project coverage, but well... For now, I think we can simply ignore the "patch" check.

PR #232, opened 2019-10-28
Update astropy-helpers package to v4.0.1

migueldvb:


mkelley:
Given APE 17, maybe this PR should be closed in favor of removing astropy-helpers, if it is possible? 
 https://github.com/astropy/astropy-APEs/blob/master/APE17.rst

migueldvb:
That's a good point. It seems that there is no rush to move away from astropy-helpers as it is going to be supported until end of 2021. The recommendation from the developers is to wait for the migration guide being finalized since it may save some time.

migueldvb:
Updating to astropy-helpers v4.0.1. If the tests pass we can merge this until we figure out how to move away from astropy-helpers.

mkelley:
@migueldvb is this good to merge?

migueldvb:
Yes, it looks good

PR #233, opened 2019-10-30
Included HG12_Pen16.evaluate method

maxmahlke:
Hi everyone!
I noticed that the `HG12_Pen16` class lacks an implementation of the `evaluate` method, so it wrongly inherits the one from the original `HG12`, which uses the 2010 definition to calculate G1 and G2 from G12. You had already implemented the changes to `fit_deriv`, it seems that `evaluate` was simply forgotten.

Before the patch:

    import numpy as np
    import sbpy.photometry as phot

    hg12 = phot.HG12()
    hg12.evaluate(ph=np.radians(10), h=10, g=1)
    # 10.533212515777848

    hg12_pen16 = phot.HG12_Pen16()
    hg12_pen16.evaluate(ph=np.radians(10), h=10, g=1)
    # 10.533212515777848

After the patch, we correctly see a difference in the `HG12_Pen16` evaluation:

    hg12_pen16 = phot.HG12_Pen16()
    hg12_pen16.evaluate(ph=np.radians(10), h=10, g=1)
    # 10.592822670157805

I would further suggest to change the variable name `g` to `g12` in the `evaluate` and `fit_deriv` calls for clarity, but didn't implement this as you might have a reason for this.


maxmahlke:
I have changed the values for `TestHG12_Pen16.test_evaluate` to the corrected G12_Pen16 model outputs.

I further changed the `g` parameter in both the `HG12` and `HG12_Pen16` in the `fit_deriv` and `evaluate` methods to `g12`.

 I see the Travis check still fails, but from what I read in the log this seems to be unrelated to these changes. @jianyangli, could you have a look at the Travis log and let me know what's going wrong? I have no experience with Travis.

mkelley:
@maxmahlke We've fixed our test environment.  I think if you rebase against master, the error will be resolved.

maxmahlke:
@mkelley @jianyangli I have rebased the PR and updated the `to_mag` and `from_obs` test methods.
The `to_ref` test is failing with an astropy unit error:
```
  File "/run/media/max/hdd/astro/repos/sbpy/sbpy/data/decorators.py", line 180, in wrapper
    return wrapped_function(*bound_args.args, **bound_args.kwargs)
  File "/run/media/max/hdd/astro/repos/sbpy/sbpy/photometry/core.py", line 704, in to_ref
    cross_section=np.pi*self.radius**2))
  File "/usr/lib/python3.7/site-packages/astropy/units/quantity.py", line 677, in to
    return self._new_view(self._to_value(unit, equivalencies), unit)
  File "/usr/lib/python3.7/site-packages/astropy/units/quantity.py", line 649, in _to_value
    equivalencies=equivalencies)
  File "/usr/lib/python3.7/site-packages/astropy/units/core.py", line 989, in to
    return self._get_converter(other, equivalencies=equivalencies)(value)
  File "/usr/lib/python3.7/site-packages/astropy/units/core.py", line 920, in _get_converter
    raise exc
  File "/usr/lib/python3.7/site-packages/astropy/units/core.py", line 906, in _get_converter
    self, other, self._normalize_equivalencies(equivalencies))
  File "/usr/lib/python3.7/site-packages/astropy/units/core.py", line 890, in _apply_equivalencies
    unit_str, other_str))
astropy.units.core.UnitConversionError: 'mag' and '1 / sr' are not convertible
```
I have not found a way to fix this, any idea?

mkelley:
@maxmahlke @jianyangli The issue here is that the solar magnitude for the `'V'` band wasn't set.  I see in the test file that it is only set in `TestDiskIntegratedPhaseFunc.test_ref_phasefunc`:
```python
solar_fluxd.set({'V': -26.77 * u.mag})
```
So, if this test isn't run, then the remaining tests that rely on having the `'V'` band flux density will fail.  To demonstrate, I can run just the tests matching 'to_ref' and they all fail:
```
$ pytest -k to_ref sbpy/photometry/tests/test_core.py
...
=============================== short test summary info ================================
FAILED sbpy/photometry/tests/test_core.py::TestLinear::test_to_ref - astropy.units.co...
FAILED sbpy/photometry/tests/test_core.py::TestHG::test_to_ref - astropy.units.core.U...
FAILED sbpy/photometry/tests/test_core.py::TestHG1G2::test_to_ref - astropy.units.cor...
FAILED sbpy/photometry/tests/test_core.py::TestHG12::test_to_ref - astropy.units.core...
FAILED sbpy/photometry/tests/test_core.py::TestHG12_Pen16::test_to_ref - KeyError: 'V'
=========================== 5 failed, 45 deselected in 2.29s ===========================
```
So, this is a bug in the tests.  The default solar flux densities could be (will be?) affected for all tests run after `photometry.test_core`, which isn't good for the rest of `sbpy`.

The fix could be adding module setup/teardown functions.  Delete line 29, where the `solar_fluxd.set` is currently called, and add to the top of the file:
```python
def setup_module(module):
    module.solar_fluxd_default = solar_fluxd.get()
    solar_fluxd.set({'V': -26.77 * u.mag})


def teardown_module(module):
    solar_fluxd.set(module.solar_fluxd_default)
```
Then the tests can pass, independent of running `TestDiskIntegratedPhaseFunc.test_ref_phasefunc`.  However, `TestHG12_Pen16.test_to_ref` will still fail, but it is because the results are different from the test values, and not because of the unit conversion.

maxmahlke:
Thanks @mkelley, I implemented your changes and updated the `to_ref` results to the expected values. The `photometry/core.py` part looks good now, the tests are failing due to a failed MPC query in ` sbpy/data/ephem.py`. I'm guessing this is unrelated to the PR.

mkelley:
Great, and yes, the MPC has had a problem with their ephemeris service and 2P/Encke (which is needed for our tests) ends up being unavailable.  I have asked them to investigate.  When they fix this, I will restart the test.

mkelley:
@jianyangli Are the changes to the test values acceptable?

jianyangli:
@maxmahlke @mkelley All the changes look good to me.  I'll merge it.  Thank you!

PR #236, opened 2020-02-07
Fix version requirements for 0.2 branch.

mkelley:
I couldn't get it to work with astropy 4, so limiting to 3.x.  Also, synphot <0.2 to be safe.  openorb now requires python 3.6 or 3.7, so upping python to 3.6 (was 3.5).

### DIFF
--- a/.rtd-environment.yml
+++ b/.rtd-environment.yml
@@ -4,9 +4,9 @@ channels:
     - astropy
 
 dependencies:
-    - python>=3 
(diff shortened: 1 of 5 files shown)


migueldvb:
The Travis error seems in the build seems to be a timeout problem.

mkelley:
Oops, pushed to the wrong branch!  I haven't learned all the VSCode tricks.

PR #238, opened 2020-02-24
better __getitem__ handling in data.DataClass

mommermi:
Currently, slicing and some other `__getitem__` calls on `DataClass` objects return `astropy.table.QTable` instances instead of objects of the same class. This commit fixes this behavior in the following way:
* if a single field name is requested (e.g., `data['RA']`), an `astropy.table.Column` object is returned (as if the method is applied to a table);
* in any other case, the method returns an instance of the same class as the current instance.

### DIFF
--- a/sbpy/data/core.py
+++ b/sbpy/data/core.py
@@ -614,17 +614,20 @@ class DataClass():
 
     def __getitem__(self, ident):
         """Return columns or rows from data table(``self._table``); checks
-        for and may use alternative field names."""
+        for and may use alternative field names. This method will always return
+        an instance of __class__, except in the case when a field name is
+        requested (then return a Table column)."""
 
+        # ignore slices
+        if isinstance(ident, slice):
+            pass
         # iterable
-        if isinstance(ident, (list, tuple, ndarray)):
+        elif isinstance(ident, (list, tuple, ndarray)):
             if all([isinstance(i, str) for i in ident]):
                 # list of column names
                 self = self._convert_columns(ident)
                 newkeylist = [self._translate_columns(i)[0] for i in ident]
                 ident = newkeylist
-                # return as new DataClass object
-                return self.from_table(self._table[ident])
             # ignore lists of boolean (masks)
             elif all([isinstance(i, bool) for i in ident]):
                 pass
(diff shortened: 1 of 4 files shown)


mkelley:
Sounds good, except that `QTable` returns `Quantity` objects, in order to carry the units.  `Column` objects don't behave the same way, which, I think, is why `QTable` was invented.  In order to preserve backwards compatibility, shouldn't single field name requests return `Quantity`?

mommermi:
Ah, yes, my fault. I only tested it without units and then getting a field from a table returns `Column`; if units are present, a `Quantity` is returned. I'll update the docs. Thanks!

sbpy/data/core.py
```diff
@@ -614,17 +614,20 @@ def __repr__(self):
...
     def __getitem__(self, ident):
         """Return columns or rows from data table(``self._table``); checks
-        for and may use alternative field names."""
+        for and may use alternative field names. This method will always return
+        an instance of __class__, except in the case when a field name is
+        requested (then return a Table column)."""
 
+        # ignore slices
+        if isinstance(ident, slice):
+            pass
```
migueldvb: Looks good.  One small thing, why are slice and int checked in the conditional statement if there is a pass and the new instance `self.from_table(self._table[ident])` will be returned?

sbpy/data/core.py
```diff
@@ -635,11 +638,12 @@ def __getitem__(self, ident):
         elif isinstance(ident, str):
             self = self._convert_columns(ident)
             ident = self._translate_columns(ident)[0]
+            return self._table[ident]
         elif isinstance(ident, int):
-            return self.from_table(self._table[ident])
+            pass
```
migueldvb: Same thing for int ident.

sbpy/data/core.py
mommermi: Those `if` statements were just for me to check on the behavior for different cases. As of bcc8538 this is condensed in a more useful structure.

sbpy/data/core.py
mommermi: ... and the same answer. Thanks for checking!

Issue #239, opened 2020-03-22
set unit error of converted CircularAperture


Yeqzids:
**High-level problem description**
CircularAperture converted to linear units lead to "set unit error" in `sbpy.activity.total_number` function. (RectanglarAperture works fine.)

**What did you do?**
```
import astropy.units as u
from sbpy.activity import gas

Q = 1e25 / u.s
v = 1 * u.km / u.s
parent = gas.photo_lengthscale('H2O')
daughter = gas.photo_lengthscale('OH')
coma = gas.Haser(Q, v, parent, daughter)

from sbpy.activity import CircularAperture
ap = CircularAperture(1 * u.arcsec).as_length(1 * u.au)
coma.total_number(ap)
```

**What did you expect?**
Number of molecules in the aperture (e.g. 2.528e+28).

**What did really happen?**
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "[..].local/lib/python3.5/site-packages/sbpy-0.2.1.dev993-py3.5.egg/sbpy/data/decorators.py", line 292, in wrapper
    **bound_args.kwargs)
  File "[..].local/lib/python3.5/site-packages/sbpy-0.2.1.dev993-py3.5.egg/sbpy/data/decorators.py", line 180, in wrapper
    return wrapped_function(*bound_args.args, **bound_args.kwargs)
  File "[..].local/lib/python3.5/site-packages/sbpy-0.2.1.dev993-py3.5.egg/sbpy/activity/gas/core.py", line 634, in total_number
    + self._K1(y) - self._K1(x)))
  File "[..]anaconda3/envs/py35/lib/python3.5/site-packages/astropy/units/quantity.py", line 648, in __array_ufunc__
    return self._result_as_quantity(result, unit, out)
  File "[..]anaconda3/envs/py35/lib/python3.5/site-packages/astropy/units/quantity.py", line 687, in _result_as_quantity
    out._set_unit(unit)
AttributeError: 'numpy.ndarray' object has no attribute '_set_unit'
```

**Provide information on your environment:**
operating system and version: [Linux (which distribution?), MacOS, Win] CentOS 7
sbpy version: 0.2.1.dev993
astropy version: 3.0.4 (note: `sbpy.activity` breaks under newer astropy; see #234 )
numpy version: 1.15.2

mkelley:
[...]

PR #240, opened 2020-03-23
Force Quantity to value in Haser (#239)

mkelley:
Force the unitless Quantity `daughter / (parent - daughter)` to a float when calculating `Haser.total_number`.  Also, favor `.to_value` over `.to(unit).value` in the code.


mkelley:
Errors are due to the MPC's ephemeris service missing 2P/Encke.  I've asked them to investigate.

PR #243, opened 2020-04-10
Add test for Haser.column_density with angular rho.

mkelley:
[skip ci]


mkelley:
The coverage report is confused because of the failed build after merging Miguel's doc changes.  I'll merge anyway.

PR #246, opened 2020-04-22
Fixed xmlns parameter in SVG file

migueldvb:
This fixes the SVG image in the website footer that I broke in #244.


mkelley:
Hmm, the PR builds are running in Travis, but they are not blocking the merges.

migueldvb:
> Hmm, the PR builds are running in Travis, but they are not blocking the merges.

That is curious. CI builds arrived but I have no idea what could have changed to allow merging without them.

mkelley:
I checked the settings, and I think the requirement was removed.  I have re-enabled it.

PR #255, opened 2020-04-30
Remove equal sign after extras_require section

migueldvb:


migueldvb:
The travis error does not seem to be related.

mkelley:
The last build on master 2 days ago was fine.  I've restarted that test.

PR #257, opened 2020-05-13
Install graphviz when testing the docs

migueldvb:
* Installing the graphviz package only when testing the docs.
* conda is overriding the pinned pytest version and tests are run currently with v5.4.3.
```
Collecting package metadata (current_repodata.json): ...working... done
Solving environment: ...working... WARNING conda.models.version:get_matcher(531): Using .* with relational operator is superfluous and deprecated and will be removed in a future version of conda. Your spec was 3.7.*, but conda is ignoring the .* and treating it as 3.7
WARNING conda.core.solve:_add_specs(600): pinned spec pytest[version='<3.7.*'] conflicts with explicit specs.  Overriding pinned spec.
done
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,14 +45,12 @@ jobs:
           env: PYTHON_VERSION=3.6
                NUMPY_VERSION=1.17
                ASTROPY_VERSION=3
-               PYTEST_VERSION='<3.7'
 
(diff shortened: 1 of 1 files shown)


mkelley:
Looks good!

PR #260, opened 2020-05-17
Add support for astropy stable

migueldvb:
`InheritDocstrings` metaclass has been dropped in astropy 4.0 so we don't need to have it in the metaclass definition for astropy>=4.


migueldvb:
This part was working locally for astropy 3 and 4, but I didn't notice
that synphot 1.3 is not supported with astropy 4.0. It seems that we
also need to change a doctest in `spectroscopy/sources.rst` (should
work now for synphot >= 2). I have changed the required versions in
the travis configuration and restarted the checks.

migueldvb:
codecov is working fine but the Horizons remote tests are still failing.

migueldvb:
This is ready to be looked at, but no rush.

mkelley:
Looks good, thanks!

PR #262, opened 2020-05-18
Separate Orbit's remote tests, and fix element testing.

mkelley:
Putting this together, I've uncovered a bug.  After orbit transformation though openorb, the orbit angles have units of deg, but based on the values, I think they should be radians.  This causes a test failure in this PR (the master code tests the values, but does not assert that they be True).  The fix for the units is TBD.


mkelley:
Not yet ready, having issues with oo_transform and radians/degrees.

mkelley:
Found the problem with oo_propagate.  pyoorb returns units of deg after propagation, but sbpy assumes all angles are in units of radians.

PR #267, opened 2020-07-24
Remove example entry point from package template

mwcraig:
There is an entry defined in `setup.cfg` that is left over from the astropy package template. This removes that entry point.


migueldvb:
Thank you!

PR #278, opened 2021-03-17
Vectorial model inclusion in sbpy.activity

sjoset:
This contains an implementation of the vectorial model based on Festou 1981 (1981A&A....95...69F), tested against the outputs of Festou's fortran code.

The VectorialModel class derives from GasComa, so is a near drop-in replacement for Haser models.
Included are two unit tests and a notebook with examples of using the model and sifting through all the wonderful things it calculates.


mkelley:
Thanks, @sjoset !  I'll check this out and get back to you within a week.

sbpy/activity/gas/core.py
```diff
@@ -635,3 +638,477 @@ def total_number(self, aper, eph=None):
...
         return N
     total_number.__doc__ = GasComa.total_number.__doc__
+
+
+class VectorialModel(GasComa):
+    """ Vectorial model for fragments in a coma produced with a dissociative energy kick
+
+        Input is a dictionary with relevant data:
+
+            --- Production Rates ---
```
mkelley, line 648: This should be converted to the numpy documentation style, e.g.,:

```

    Parameters
    ----------
    Q : `~astropy.units.Quantity`
        Production rate, per time, may be an array with length equal to that of ``dt``.
    dt : `~astropy.units.Quantity`
       Relative time with respect to now (e.g., model reference/observation time), with positive numbers indicating the past.
    ...etc...
```

sbpy/activity/gas/core.py
```diff
@@ -635,3 +638,477 @@ def total_number(self, aper, eph=None):
...
+    """ NOTE: We allow Q to vary with time, so we pass it along to the base class but it is ultimately ignored in favor
+        of the production information coming in through inputParameters.  Likewise, v is also ignored because the parent
+        outflow velocity is included in inputParameters.
+        Calculations are done internally in meters and seconds to match the base GasComa class.
+    """
+    # TODO: We would have to rewrite the time-varying production stuff to accept Q here and then let the param structure
+    # define some variation of this initial production rate
+    @bib.cite({'model': '1981A&A....95...69F'})
+    @u.quantity_input(Q=(u.s**-1, u.mol / u.s), v=u.m / u.s)
+    def __init__(self, Q, v, inputParameters):
```
mkelley, line 696: There are a lot of options, but I think we can unpack them from the dictionary and leave them as individual keyword arguments.  Let me propose the following:

```
def __init__(self, Q, dt, parent, fragment, radial_points=X, radial_substeps=Y, angular_points=U, angular_substeps=V):
```
Then `parent` and `fragment` can use our ``Phys`` data class:
```python
parent = Phys.from_dict({
    tau_T=100 * u.day,
    tau_d=4e5 * u.s,
    v=1 * u.km / u.s,
    sigma=1e-10 * u.cm**2
})
```
Destruction level sounds like a model parameter, rather than a physical property, so perhaps it should also be a keyword argument to `__init__()`.

sbpy/activity/gas/core.py
```diff
@@ -635,3 +638,477 @@ def total_number(self, aper, eph=None):
...
+        # Convert to seconds
+        self.vModelParams['TimeAtProductions'] = self.vModelParams['TimeAtProductions'].to(u.s).value
+        self.numProductionRates = len(self.vModelParams['ProductionRates'])
+
+        # Make sure we have as many production times as rates
+        if self.numProductionRates != len(self.vModelParams['TimeAtProductions']):
+            raise ValueError("Number of production rates and production times must match!")
+
+        # fill a few short names for readability later
+        self.par = self.vModelParams['Parent']
```
mkelley, line 710: Prefer spelling out `self.parent` and `self.fragment`.

sbpy/activity/gas/core.py
```diff
@@ -635,3 +638,477 @@ def total_number(self, aper, eph=None):
...
+        if 'Grid' not in self.vModelParams:
+            self.vModelParams['Grid'] = {}
+        if 'NumRadialGridpoints' not in self.vModelParams['Grid']:
+            self.vModelParams['Grid']['NumRadialGridpoints'] = 50
+        if 'NumAngularGridpoints' not in self.vModelParams['Grid']:
+            self.vModelParams['Grid']['NumAngularGridpoints'] = 25
+        if 'NumSubgridRadialSteps' not in self.vModelParams['Grid']:
+            self.vModelParams['Grid']['NumSubgridRadialSteps'] = 12
+        if 'NumSubgridAngularSteps' not in self.vModelParams['Grid']:
+            self.vModelParams['Grid']['NumSubgridAngularSteps'] = 7
```
mkelley, line 746: These defaults would be moved to the `__init__()` line.

sbpy/activity/gas/core.py
```diff
@@ -635,3 +638,477 @@ def total_number(self, aper, eph=None):
...
+        if(vf < vp):
+            self.vModel['EpsilonMax'] = np.arcsin(vf/vp)
+        else:
+            self.vModel['EpsilonMax'] = np.pi
+
+    def productionRateAtTime(self, t):
+        """ Get production rate at time t in s, with positive values representing the past, so
+            productionRateAtTime(5) will provide the production 5 seconds ago
+
+            For times in the past beyond the specified production rates, returns zero
```
mkelley, line 829: Use numpy doc format.

sbpy/activity/gas/core.py
```diff
@@ -635,3 +638,477 @@ def total_number(self, aper, eph=None):
...
+
+        self.vModel['ColumnDensity'] = {}
+        self.vModel['ColumnDensity']['FastCDGrid'] = cDensGrid
+        self.vModel['ColumnDensity']['CDGrid'] = cDensGrid*u.m
+        self.vModel['ColumnDensity']['Values'] = columnDensities/(u.m**2)
+        # Interpolator gives column density in m^-2
+        self.vModel['ColumnDensity']['Interpolator'] = CubicSpline(cDensGrid, columnDensities, bc_type='natural')
+
+    def calcNumFragmentsTheory(self):
+        """ Returns the total number of fragment species we expect in the coma theoretically """
```
mkelley, line 1066: Use numpy doc format.  For methods that return values:
```
    Returns
    -------
    theoryTot : float
        Total number of fragment species.
```

sbpy/activity/gas/core.py
```diff
@@ -635,3 +638,477 @@ def total_number(self, aper, eph=None):
...
+                t2 = 0
+            theoryTot += pRates[i]*(-np.e**(-t1) + np.e**(-t2))
+
+        theoryTot = theoryTot*(fTotLifetime*pTotLifetime/pDisLifetime) \
+            - (np.pi * mR * mR * (vf + vp) * self.vModel['FastRadialDensity'][lastDensityElement])
+
+        return theoryTot
+
+    def calcNumFragmentsFromGrid(self):
+        """ Returns the total number of fragment species by integrating the density grid over the volume """
```
mkelley, line 1098: Use numpy doc format.

mkelley (commented):
Thanks, @sjoset , and apologies for the long delay.  I've made some initial comments.  We also need to rebase with the current master branch so that we can get the code checks to work.  Do you want to implement the changes or leave them to us?

sjoset:
Hi Dr. Kelley,

I've implemented the changes to the vectorial model initialization you suggested, switched to numpy doc format, ironed out the PEP8 issues as best as possible, and updated the example notebook & unit tests.
I'm new to pull requests - do I need to make another, or are the updates to my fork included?

Also, thank you for the suggestions!  They worked out nicely.

Thanks,
Shawn
________________________________

sbpy/activity/gas/tests/test_core.py
```diff
@@ -392,3 +393,60 @@ def test_missing_scipy(self, monkeypatch):
             test._iK0(1)
         with pytest.raises(sbe.RequiredPackageUnavailable):
             test._K1(1)
+
+class TestVectorialModel:
```
mkelley, line 397: Could a test be developed that reproduces results from Festou's second paper?  https://ui.adsabs.harvard.edu/abs/1981A%26A....96...52F/abstract

I see some quoted values in Section IV that might help develop a test.

mkelley (changes requested):
The code looks good.  It is well-formatted and the comments and docstrings are really helpful.  I've just added a few small requests.  However, the tests are failing for me so please address those failures.

The errors are due to the parent and fragment molecules as `Phys` objects: `Field v does not have units of rad`.  We've implemented unit checks on `Phys` objects, and it expects an angle.  This is because 'v' is already used for true anomaly in `Orbit`.  Can you recommend other variable names?  Perhaps 'v_outflow' and 'v_photo'?

Also, there are some code style failures on the GitHub checks.  I'll quote them here for convenience:
sbpy/activity/gas/core.py:741:17: E241 multiple spaces after ':'
sbpy/activity/gas/core.py:748:17: E241 multiple spaces after ':'
sbpy/activity/gas/tests/test_core.py:397:1: E302 expected 2 blank lines, found 1

sbpy/activity/gas/tests/test_core.py
mkelley: If this is too tricky, we can leave it as an open issue.

PR #283, opened 2021-05-28
Fix different results between model evaluation and direct call to `.evaluate` method for photometry models

jianyangli:


jianyangli:
Closed and replaced by #286

PR #284, opened 2021-06-17
Infrastructure update

mkelley:
Following the migration guide to dropping astropy_helpers and modernize the testing infrastructure

https://docs.astropy.org/projects/package-template/en/latest/ape17.html

tox is running! although not all test environments pass.

Unfortunately, I've had to drop pyoorb from the installation requirements because it is not pip installable.  ~As a result, tox will not run any pyoorb tests.~  This may be resolved if https://github.com/mkelley/pyoorb-experiment can be used.

Next step: migrate to CI to github actions and drop travis.


mkelley:
@jianyangli This PR should get our testing working so that we can merge branches again.

PR #286, opened 2021-06-20
bugfix in `sbpy.photometry`

jianyangli:
Replace PR #283

Fix different results between model evaluation and direct call to .evaluate method for photometry models

PR #288, opened 2021-06-20
bugfix in `sbpy.photometry`

jianyangli:
This PR replaces #286 , which replaced PR #283

Fix different results between model evaluation and direct call to .evaluate method for photometry models

### DIFF
--- a/sbpy/photometry/core.py
+++ b/sbpy/photometry/core.py
@@ -1026,8 +1026,7 @@ class HG1G2(HG12BaseClass):
 
     @staticmethod
     def evaluate(ph, h, g1, g2):
-        if isinstance(ph, u.Quantity):
-            ph = ph.to('rad').value
+        ph = u.Quantity(ph, 'deg').to_value('rad')
         func = g1*HG1G2._phi1(ph)+g2*HG1G2._phi2(ph)+(1-g1-g2)*HG1G2._phi3(ph)
         if isinstance(func, u.Quantity):
             func = func.value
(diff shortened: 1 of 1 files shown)


jianyangli:
@anttipenttila Can you verify that this fix works for you?  I'll merge it once hear back from you.  Thanks.

sbpy/photometry/core.py
```diff
@@ -1026,6 +1026,8 @@ def _G2(self):
 
     @staticmethod
     def evaluate(ph, h, g1, g2):
+        if isinstance(ph, u.Quantity):
+            ph = ph.to('rad').value
```
mkelley, line 1030: Where is the default behavior (assumed units on `ph`) documented?  Would `ph = u.Quantity(ph, 'deg').to_value('rad')` be acceptable?  That way the default (if it is degrees) is clear in the code and the if statement can be avoided.

jianyangli: Good trick.  Will do it.  Thanks.

mkelley:
Hi @jianyangli @anttipenttila, can this be resolved soon?  Thanks.

jianyangli:
@anttipenttila, I will merge this branch by next Tuesday, July 13.  If I don't hear from you by then, then we can revisit it later if you find other bugs.  Thanks.

anttipenttila:
You can merge, I'm sure you tested with the example I had that showed the bug, and I don't have anything else on this.

PR #281, opened 2021-05-23
add spectral reddening

jianyangli:
Add spectral reddening functionality to `~spectroscopy.SpectralSource` class

PR #289, opened 2021-06-20
Add spectral reddening

jianyangli:
This PR replaces #281 

Add spectral reddening functionality to `sbpy.spectroscopy.SpectralSource` class

### DIFF
--- a/docs/sbpy/spectroscopy/index.rst
+++ b/docs/sbpy/spectroscopy/index.rst
@@ -58,6 +58,56 @@ Renormalize to 1.0 μm:
   <SpectralGradient 6.89655172 % / 100 nm>
 
 
+Spectral Reddening
+------------------
+
+Linear spectral reddening is enabled by class `~sbpy.spectroscopy.Reddening`,
+which is based on `~synphot.BaseUnitlessSpectrum`.
+
+Initialize a `~sbpy.spectroscopy.Reddening` class from a spectral gradient:
+
+
+.. plot::
+  :include-source: True
+
+  import matplotlib.pyplot as plt
+  import astropy.units as u
+  from sbpy.spectroscopy import SpectralGradient, Reddening
+  from sbpy.units import hundred_nm
+
+  S = SpectralGradient(10 * u.percent / hundred_nm, wave0=0.55 * u.um)
+  linear_reddening = Reddening(S)
+  linear_reddening.plot()
+  plt.gca().grid()
+
+
+This class can then be used to linearly redden a spectrum as a
+`~synphot.SourceSpectrum` class instance:
+
+
+.. plot::
+  :include-source: True
+
+  import numpy as np
+  import matplotlib.pyplot as plt
+  import astropy.units as u
+  from synphot import SourceSpectrum
+  from synphot.models import BlackBodyNorm1D
+  from sbpy.spectroscopy import SpectralGradient, Reddening
+  from sbpy.units import hundred_nm
+
+  S = SpectralGradient(10 * u.percent / hundred_nm, wave0=0.55 * u.um)
+  linear_reddening = Reddening(S)
+  spec = SourceSpectrum(BlackBodyNorm1D, temperature=5500 * u.K)
+  reddened = spec * linear_reddening
+  wv = np.linspace(0.3, 1, 100) * u.um
+
+  plt.plot(wv, spec(wv))
+  plt.plot(wv, reddened(wv))
+  plt.legend(['Original', 'Reddened'])
+  plt.setp(plt.gca(), xlabel='Wavelength (um)', ylabel='Flux (PHOTLAM)')
+  plt.tight_layout()
+
 
 Reference/API
 -------------
--- a/docs/sbpy/spectroscopy/index.rst
+++ b/docs/sbpy/spectroscopy/index.rst
@@ -62,7 +62,7 @@ Spectral Reddening
 ------------------
 
 Linear spectral reddening is enabled by class `~sbpy.spectroscopy.Reddening`,
-which is based on `~synphot.BaseUnitlessSpectrum`.
+which is based on `~synphot.spectrum.BaseUnitlessSpectrum`.
 
 Initialize a `~sbpy.spectroscopy.Reddening` class from a spectral gradient:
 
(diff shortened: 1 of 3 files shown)


sbpy/spectroscopy/sources.py
```diff
@@ -467,6 +467,29 @@ def color_index(self, wfb, unit):
...
+            The spectral gradient to redden.
+
+        Returns
+        -------
+        spec : `~SpectralSource`
+            Reddened spectrum
+
+        """
+        r = Reddening(S)
+        from copy import deepcopy
```
mkelley, line 485: Move this import to the top.

sbpy/spectroscopy/sources.py
```diff
@@ -498,3 +521,19 @@ def __repr__(self):
...
+    Parameters
+    ----------
+    S : `~SpectralGradient`
+        The spectral gradient to redden.
+    """
+    def __init__(self, S):
+        wv = [1, 2] * S.wave0
+        df = (S.wave0 * S).to('').value
+        super().__init__(
+            synphot.Empirical1D, points=wv, lookup_table=[1, 1+df],
```
mkelley, line 538: I don't see a Linear1D model in synphot, so I think Empirical1D is the right choice.  The only problem might be extrapolation beyond the two wavelength points, for which synphot has several options.  So what can we do to avoid or mitigate that.  We can certainly provide a data point for `wavelength = 0`.  Maybe also a very large value, e.g., 1 cm or `100 * wave0`?

mkelley:
Also, please add a new section to the documentation in `sbpy/spectroscopy/index.rst`

sbpy/spectroscopy/sources.py
jianyangli: By setting `fill_value=None` to `Empirical1D`, the model will extrapolate outside of the supplied wavelength range.  When only two points are supplied, the extrapolation is linear.  This way, we can avoid having to provide 0 and a very large value for wavelength.

mkelley: You're right, it does work.  I was expecting a synphot DisjointError when trying to observe the result with, e.g., a filter that does not overlap the wavelength range specified here, but that doesn't seem to be an issue.

jianyangli:
I have one general question about `sbpy.spectroscopy.sources.SpectralSource`.  I wonder why we don't directly subclass `synphot.SourceSpectrum`, but rather store its instance in `SpectralSource`.  Direct subclass will allow various operations defined for `synphot.SourceSpectrum`, such as add, subtraction, and multiplications of two spectra.  This would simplify some operations such as spectral reddening.  Thoughts?

mkelley:
I agree with those benefits.  I think I assumed we would using a Spectral1D class derived from specutils at some point.  It's worth revisiting that assumption and considering the merits of each.

jianyangli:
> I agree with those benefits. I think I assumed we would using a Spectral1D class derived from specutils at some point. It's worth revisiting that assumption and considering the merits of each.

OK, then I will submit an issue to document this idea for future work.  Thanks.

docs/sbpy/spectroscopy/index.rst
```diff
@@ -58,6 +58,56 @@ Renormalize to 1.0 μm:
   <SpectralGradient 6.89655172 % / 100 nm>
 
 
+Spectral Reddening
+------------------
+
+Linear spectral reddening is enabled by class `~sbpy.spectroscopy.Reddening`,
+which is based on `~synphot.BaseUnitlessSpectrum`.
```
mkelley, line 65: To get this to link, use `~synphot.spectrum.BaseUnitlessSpectrum`

mkelley (approved):
Since many users will want to redden a solar spectrum, can you also add this example:

```
sbpy ``SpectralSource`` objects have a ``redden()`` method for reddening.  The following example reddens a solar spectrum:
 
.. plot::
  :include-source: True

  import matplotlib.pyplot as plt
  import astropy.units as u
  from sbpy.calib import Sun
  from sbpy.spectroscopy import SpectralGradient, Reddening
  from sbpy.units import hundred_nm

  sun = Sun.from_builtin('E490_2014LR') # low-resolution solar spectrum
  S = SpectralGradient(10 * u.percent / hundred_nm, wave0=0.55 * u.um)
  red_sun = sun.redden(S)

  wv = np.linspace(0.3, 1, 100) * u.um
  fluxd_unit = u.Unit('W/(m2 um)')

  plt.plot(wv, sun.observe(wv, unit=fluxd_unit))
  plt.plot(wv, red_sun.observe(wv, unit=fluxd_unit))
  plt.legend(['Original', 'Reddened'])
  plt.setp(plt.gca(), xlabel='Wavelength (um)', ylabel='Flux density ({})'.format(fluxd_unit))
  plt.tight_layout()

```

The SpectralSource object doesn't seem to be documented, maybe because it is an abstract class.  I think we can address that when we reconsider the sub-classing.

jianyangli:
> Since many users will want to redden a solar spectrum, can you also add this example:
> 
> ```
> sbpy ``SpectralSource`` objects have a ``redden()`` method for reddening.  The following example reddens a solar spectrum:
>  
> .. plot::
>   :include-source: True
> 
>   import matplotlib.pyplot as plt
>   import astropy.units as u
>   from sbpy.calib import Sun
>   from sbpy.spectroscopy import SpectralGradient, Reddening
>   from sbpy.units import hundred_nm
> 
>   sun = Sun.from_builtin('E490_2014LR') # low-resolution solar spectrum
>   S = SpectralGradient(10 * u.percent / hundred_nm, wave0=0.55 * u.um)
>   red_sun = sun.redden(S)
> 
>   wv = np.linspace(0.3, 1, 100) * u.um
>   fluxd_unit = u.Unit('W/(m2 um)')
> 
>   plt.plot(wv, sun.observe(wv, unit=fluxd_unit))
>   plt.plot(wv, red_sun.observe(wv, unit=fluxd_unit))
>   plt.legend(['Original', 'Reddened'])
>   plt.setp(plt.gca(), xlabel='Wavelength (um)', ylabel='Flux density ({})'.format(fluxd_unit))
>   plt.tight_layout()
> ```
> 
> The SpectralSource object doesn't seem to be documented, maybe because it is an abstract class. I think we can address that when we reconsider the sub-classing.

OK, this is added now.  Thanks.

mkelley:
Excellent!

PR #293, opened 2021-06-28
Fixing weekly tests

mkelley:
Add missing name in workflow step?


mkelley:
Works offline... let's try on.

Issue #303, opened 2021-10-13
GasComa.total_number does not accept Quantity


mkelley:
**High-level problem description**
GasComa.total_number does not accept Quantity for apertures, but documentation says it should.

This came about by testing #278.

```
        coma = VectorialModel(base_q=Q0,
                              parent=parent,
                              fragment=fragment)
        coma.total_number(10 * u.arcsec, eph=delta)
```

```
sbpy/activity/gas/core.py:344: in total_number
    aper = aper.as_length(eph)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <Quantity 10. arcsec>, attr = 'as_length'

    def __getattr__(self, attr):
        """
        Quantities are able to directly convert to other units that
        have the same physical type.
        """
        if not self._include_easy_conversion_members:
>           raise AttributeError(
                f"'{self.__class__.__name__}' object has no '{attr}' member")
E           AttributeError: 'Quantity' object has no 'as_length' member

/home/msk/.venv/lib/python3.8/site-packages/astropy/units/quantity.py:848: AttributeError
```

PR #306, opened 2021-10-20
Aper can be a quantity; re-using code preferred.

mkelley:
Address issue #303 and edit a few lines to re-use code by keeping aperture normalization in ``GasComa.total_number`` and move actual calculation to ``_total_number``.


jianyangli (approved):
All changes look good to me in terms of coding, although I'm not very familiar with the physical model implemented here.

mkelley:
Thanks!  Might it be time to expand your cometary horizons? :)

PR #309, opened 2022-01-11
Fix linkcheck 20220111

mkelley:
Address failing weekly test:

```
Warning, treated as error:
/home/runner/work/sbpy/sbpy/docs/sbpy/spectroscopy/index.rst:70:Exception occurred in plotting index-1
 from /home/runner/work/sbpy/sbpy/docs/sbpy/spectroscopy/index.rst:
Traceback (most recent call last):
  File "/home/runner/work/sbpy/sbpy/.tox/linkcheck/lib/python3.8/site-packages/matplotlib/sphinxext/plot_directive.py", line 517, in _run_code
    exec(code, ns)
  File "<string>", line 6, in <module>
  File "/home/runner/work/sbpy/sbpy/.tox/linkcheck/lib/python3.8/site-packages/sbpy/spectroscopy/core.py", line 102, in __new__
    S = super().__new__(cls, value, unit=unit, dtype=dtype,
  File "/home/runner/work/sbpy/sbpy/.tox/linkcheck/lib/python3.8/site-packages/astropy/units/quantity.py", line 430, in __new__
    return np.array(value, dtype=dtype, copy=copy, order=order,
ValueError: NoneType copy mode not allowed.
```

And, catch an issue when doctests are run twice in a row.  obs.to_file example writes a file, but raises an error when the file already exists.

### DIFF
--- a/docs/sbpy/data/dataclass.rst
+++ b/docs/sbpy/data/dataclass.rst
@@ -576,7 +576,15 @@ Writing object data to a file
 `~sbpy.data.DataClass` objects can be written to files using
 `~sbpy.data.DataClass.to_file`:
 
-    >>> obs.to_file('observations.dat')  # doctest: +SKIP
+    .. testsetup:: obs.to_file
+        >>> import os
+        >>> assert not os.path.exists('observations.dat')
+
+    .. doctest:: obs.to_file
+        >>> obs.to_file('observations.dat')
+
+    .. testcleanup:: obs.to_file
+        >>> os.unlink('observations.dat')
 
 By default, the data are written in ASCII format, but other formats
 are available, too (`list of file formats
--- a/docs/sbpy/data/dataclass.rst
+++ b/docs/sbpy/data/dataclass.rst
@@ -576,7 +576,7 @@ Writing object data to a file
 `~sbpy.data.DataClass` objects can be written to files using
 `~sbpy.data.DataClass.to_file`:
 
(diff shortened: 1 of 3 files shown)


docs/sbpy/data/dataclass.rst
```diff
@@ -576,7 +576,15 @@ Writing object data to a file
...
-    >>> obs.to_file('observations.dat')
+    .. testsetup:: obs.to_file
+        >>> import os
+        >>> assert not os.path.exists('observations.dat')
+
+    .. doctest:: obs.to_file
+        >>> obs.to_file('observations.dat')
+
+    .. testcleanup:: obs.to_file
+        >>> os.unlink('observations.dat')
```
migueldvb, line 587: Nice solution. I would have thought you need to import os under testcleanup but apparently it is not necessary since all the checks have passed.

migueldvb (approved):
Looks great! Thanks

docs/sbpy/data/dataclass.rst
mkelley: I think that has something to do with the block name following the directives.  Here, I called the block obs.to_file.

mkelley:
Thanks, @migueldvb !

docs/sbpy/data/dataclass.rst
mkelley: Hmmm, I guess the groups are only used inside of sphinx.

PR #310, opened 2022-01-12
Restore all tests in dataclass.rst

mkelley:
Design all tests in dataclass.rst for astropy>=5.  Use conditional statements to skip tests when astropy<5 is used.

### DIFF
--- a/docs/sbpy/data/dataclass.rst
+++ b/docs/sbpy/data/dataclass.rst
@@ -166,6 +166,8 @@ propagate this orbit and obtain ephemerides. Since you are dealing
 with a single orbit, the most convenient solution might be to use a
 dictionary to build your object:
 
+.. doctest-requires:: astropy>=5
+
     >>> from sbpy.data import Orbit
     >>> from astropy.time import Time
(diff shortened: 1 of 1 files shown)


docs/sbpy/data/dataclass.rst
```diff
@@ -402,37 +431,50 @@ object:
...
      10.25546  -12.3946 2451523.94653      i       i
 
 Similarly, existing columns can be modified using:
 
-    >>> obs['filter'] = ['g', 'i', 'R', 'V']  # doctest: +SKIP
-    
+.. testsetup::
+
+    >>> from astropy.table import Column
+    >>> if len(obs) < 4:  # then this is astropy<5
```
migueldvb, line 470: Here obs is modified for astropy<5. I wonder why not having doctest-requires directive to make this dependency more explicit as in other places but perhaps there is a good reason to do this that I have missed.

migueldvb (approved):
Great work to update these tests and keep support for older astropy versions. All the checks are passing and this looks ready to go.

Issue #317, opened 2022-02-04
Inconsistent time order in Ephem.from_*


mkelley:
**High-level problem description**
When requesting ephemerides at specific epochs, the order of the returned rows may be in sorted order rather than in the same order as the original request.

**What did you do?**
```python
import numpy as np
from astropy.time import Time
from sbpy.data import Ephem

t = Time(['2022-01-01', '2021-01-01'])
mpc = Ephem.from_mpc('C/2019 L3', epochs=t)
horizons = Ephem.from_horizons('C/2019 L3', epochs=t)
miriade = Ephem.from_miriade('C/2019 L3', 'comet', epochs=t)
for eph in [mpc, horizons, miriade]:
    print(np.diff(eph['date']))
print('original:', np.diff(t))
```
Results:
```
[<TimeDelta object: scale='tai' format='jd' value=-365.0>]
[<TimeDelta object: scale='tai' format='jd' value=365.0>]
[<TimeDelta object: scale='tai' format='jd' value=-365.0>]
original: [<TimeDelta object: scale='tai' format='jd' value=-365.0>]
```

**What did you expect?**
I expected the date to be in the same order as my original array.

**What did really happen?**
The JPL Horizons result was sorted, but the others were not.  I think this is a feature of the new Horizons API.

**Provide information on your environment:**
astropy==4.3.post1
astroquery==0.4.5
sbpy==0.2.3.dev195+g99560e4

mkelley:
I think the solution is to sort epochs before sending to Horizons, then unsort the result to get back to the user's requested order.

PR #318, opened 2022-02-04
Sort and unsort epochs for JPL Horizons.

mkelley:
Addresses #317 

~Still need PR specific tests.~

### DIFF
--- a/sbpy/data/ephem.py
+++ b/sbpy/data/ephem.py
@@ -136,7 +136,7 @@ class Ephem(DataClass):
                 warn(('converting {} epochs to utc for use in '
                       'astroquery.jplhorizons').format(epochs.scale),
                      TimeScaleWarning)
-            _epochs = epochs.utc.jd
+            _epochs = epochs.utc.jd  # ~1 ms precision for modern dates
 
             # 2022 Feb 04: JPL Horizons API v1.1 will return sorted dates, but
             # we want to preserve the user's requested order.  Sort them, send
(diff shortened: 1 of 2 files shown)


sbpy/data/ephem.py
```diff
@@ -127,14 +128,26 @@ def from_horizons(cls, targetids, id_type='smallbody',
...
+        _sorted = False
         if epochs is None:
             _epochs = [Time.now().utc.jd]
         elif isinstance(epochs, Time):
             if epochs.scale != 'utc':
                 warn(('converting {} epochs to utc for use in '
                       'astroquery.jplhorizons').format(epochs.scale),
                      TimeScaleWarning)
-            _epochs = epochs.utc.jd
+            _epochs = epochs.utc.jd  # might this lose some precision?
```
jianyangli, line 139: I don't think this will lose precision.  [Astropy.time documentation](https://docs.astropy.org/en/stable/time/index.html) says`Time` will maintain sub-ns precision.  You can even specify 16-byte long-double type for output.

mkelley: Long-double output would be great, if it propagates through to the Horizons API, and if Horizons can retain that precision on their end.  Seems like we should have an open issue to investigate this.

jianyangli: Agreed.  I can verify the 1 ms precision in my computer, too.  I'm fine with just using the default (double-precision) here.

jianyangli (approved):
Everything looks good to me.

mkelley:
Thanks, @jianyangli

PR #319, opened 2022-02-04
Fix remote tests and from_mpc bugs

mkelley:
Get remote tests working by:
* updating some remote data that have changed (Ceres photometric model, again)
* update LTE.from_Drahus test to match astropy v5 behavior, and skip the test when astropy < 5
* instantiate astroquery's MPC class before using it ( astropy/astroquery#2280 )

### DIFF
--- a/sbpy/activity/gas/productionrate.py
+++ b/sbpy/activity/gas/productionrate.py
@@ -11,6 +11,7 @@ created on June 26, 2019
 
 import tempfile
 import numpy as np
+import astropy
 import astropy.constants as con
 import astropy.units as u
 from astroquery.jplspec import JPLSpec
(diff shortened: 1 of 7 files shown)


sbpy/activity/gas/productionrate.py
```diff
@@ -590,7 +595,7 @@ def from_Drahus(self, integrated_flux, mol_data, ephemobj, vgas=1 * u.km/u.s,
         ...                     ephemobj, vgas, aper, b=b) # doctest: +REMOTE_DATA
 
         >>> q  # doctest: +REMOTE_DATA +FLOAT_CMP
-        <Quantity 1.09899965e+25 1 / s>
+        <MaskedQuantity 1.09899965e+25 1 / s>
```
migueldvb, line 598: Interesting! I'm trying to figure out why this change happened in astropy 5 and to make it work also for older astropy versions. I'll make it a separate PR because these changes fix the failing remote tests and should be merged.

migueldvb (approved):
Nice work. Thanks

sbpy/activity/gas/productionrate.py
mkelley: Sounds good!

PR #325, opened 2022-04-27
Add orbit dynamics calculations

jianyangli:
Add `tisserand` and `d_criterion` methods to `Orbit` class.

### DIFF
--- a/docs/sbpy/data/orbit.rst
+++ b/docs/sbpy/data/orbit.rst
@@ -106,7 +106,7 @@ terms of Julian date. The following example propagates the current
 orbit of Ceres back to year 2000:
 
     >>> elem = Orbit.from_horizons('Ceres')  # doctest: +REMOTE_DATA
-    >>> epoch = Time('2000-01-01')
+    >>> epoch = Time('2000-01-01', scale='tdb')
     >>> newelem = elem.oo_propagate(epoch) # doctest: +SKIP 
     >>> newelem # doctest: +SKIP
     <QTable length=1>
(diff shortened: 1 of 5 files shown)


jianyangli:
I compared the numbers against references, but @mkelley can you check my implementation of the formula, especially for d_criterion, to make sure they are correct.  Thanks.

sbpy/data/tests/test_orbit.py
```diff
@@ -161,3 +162,96 @@ def test_oo_propagate(self):
...
+}
+
+
+class TestOrbitTisserandDCriterion:
+    def testTisserand(self):
+        """ test tisserand method"""
+
+        halley = Orbit.from_dict(HALLEY)
+        jupiter = Orbit.from_dict(JUPITER)
+        assert u.isclose(halley.tisserand(jupiter), u.Quantity(-0.61659744))
```
mkelley, line 242: JPL SBDB has -0.605, which we can reproduce if we use the same epoch as their orbital elements:
```python
>>> epoch = Time(2449400.5, format='jd')
>>> halley = Orbit.from_horizons('1P', id_type='designation', closest_apparition=True, epochs=epoch)
>>> jupiter = Orbit.from_horizons(599, id_type=None, epochs=epoch)
>>> halley.tisserand(jupiter)
Out[47]: <MaskedQuantity -0.60495016>
```

sbpy/data/orbit.py
```diff
@@ -724,3 +724,72 @@ def oo_propagate(self, epochs, dynmodel='N', ephfile='de430'):
...
+
+        a_p = planet['a']
+        t = a_p / self['a'] + 2 * np.cos(self['i']) * \
+            np.sqrt((1 - self['e']**2) * self['a'] / a_p)
+        t = u.Quantity(t, '')
+        if len(t) == 1:
+            t = t[0]
+        return t
+
+    @cite({'method': '1993Icar..106..603J '})
```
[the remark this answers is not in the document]
jianyangli: I only see D_SH and D_D in Jopek 1993.  Should we add an optional parameter to return D_D, too?

sbpy/data/orbit.py
```diff
@@ -724,3 +724,72 @@ def oo_propagate(self, epochs, dynmodel='N', ephfile='de430'):
...
+    def d_criterion(self, obj):
+        """Evaluate orbit similarity d-criterion
```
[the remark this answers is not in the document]
jianyangli: OK

sbpy/data/orbit.py
```diff
@@ -724,3 +724,72 @@ def oo_propagate(self, epochs, dynmodel='N', ephfile='de430'):
...
+        a_p = planet['a']
+        t = a_p / self['a'] + 2 * np.cos(self['i']) * \
+            np.sqrt((1 - self['e']**2) * self['a'] / a_p)
+        t = u.Quantity(t, '')
+        if len(t) == 1:
+            t = t[0]
+        return t
+
+    @cite({'method': '1993Icar..106..603J '})
+    def d_criterion(self, obj):
```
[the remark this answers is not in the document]
jianyangli: I'm fine with capital D in the method name.

sbpy/data/tests/test_orbit.py
```diff
@@ -161,3 +162,96 @@ def test_oo_propagate(self):
...
+        comets = Orbit.from_dict(COMETS)
+        assert u.allclose(comets.tisserand(jupiter),
+                          u.Quantity([2.82130706, 2.79709686]))
+
+    def testDCriterion(self):
+        """ test d_criterion method"""
```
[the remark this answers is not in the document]
jianyangli: I will implement the test against Table 1 of Josep 1993.

sbpy/data/tests/test_orbit.py
jianyangli: I will add this test to test_orbit_remote.py.

sbpy/data/orbit.py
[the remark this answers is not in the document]
mkelley: D_H is equation 11.  Maybe we should get the opinions of some people who actually use these equations before expanding this?

sbpy/data/orbit.py
```diff
@@ -757,24 +760,35 @@ def tisserand(self, planet):
...
+
         Returns
         -------
         float or numpy.ndarray
 
+
+        References
+        ----------
+        Southwarth, R. B., & Hawkins, G. S. 1963, SCoA, 7, 261
+        Josep, T. J. 1993, Icarus 106, 60
```
mkelley, line 782: "Josep" --> "Jopek", "60" --> "603"

jianyangli: Oops, sorry for the silly mistake.

sbpy/data/orbit.py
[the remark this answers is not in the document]
jianyangli: Got it.  Thanks.

OK, I'll consult Quan-Zhi (as a meteor person) first and see what he says.  Then we can also see the feedback from users to decide whether to expand it.

Yeqzids:
Yes, it would be good to include different versions of $D$ if possible, since each has its own caveat and it is common to calculate/compare different $D$s. A complete summary of different versions of $D$ can be found in https://ui.adsabs.harvard.edu/abs/2019msme.book..210W/abstract, Section 9.2.1.

jianyangli:
> Yes, it would be good to include different versions of $D$ if possible, since each has its own caveat and it is common to calculate/compare different $D$s. A complete summary of different versions of $D$ can be found in https://ui.adsabs.harvard.edu/abs/2019msme.book..210W/abstract, Section 9.2.1.

OK, Thanks @Yeqzids .  I will add the other versions.

jianyangli:
@mkelley , I have a question for you.  After implementing all three versions of D-criterion, I added all three references, Southworth & Hawkins 1963, Drummond 1981, and Jopek 1993 to bib track.  But ideally, we only want to track the original reference for each corresponding version of D.  E.g., if `version='d'`, then only the Drummond 1981 reference should be tracked.  Is there a way to do this?  If not, what's approach would you suggest for this case?

mkelley:
Move the bib tracking calls to inside the method and place each within the their respective if-statement blocks.

jianyangli:
@mkelley Is this bib tracking implementation OK?  Thanks.

sbpy/data/orbit.py
```diff
@@ -865,10 +864,12 @@ def D_criterion(self, obj, version='sh'):
 
             if version.lower() == 'sh':
                 # Southworth-Hawkins function
+                register(self.D_criterion, {'method': '1963SCoA....7..261S'})
                 d2 = diff_e**2 + diff_q**2 + 4 * sin_i2 \
                     + (sum_e * np.sin(pi_ba / 2))**2
             else:
```
mkelley, line 870: Can this method raise an error if `version` is not valid?

mkelley:
Yes, I think this is right.

docs/sbpy/data/orbit.rst
```diff
@@ -118,3 +118,54 @@ orbit of Ceres back to year 2000:
 
 Note that both functions require `pyoorb
 <https://github.com/oorb/oorb/tree/master/python>`_ to be installed.
+
+
+Calculate dynamic parameters
```
mkelley, line 123: dynamic --> dynamical

mkelley (approved):
After those last two comments, I think this is good!

sbpy/data/orbit.py
jianyangli: Yes it does.  Lines 827 - 829 checks the value of `version` and return an exception if not one of the three.

This reminds me:  Do we have a convention in `sbpy` to raise `SbpyException` for errors?  Right now it raises a generic `ValueError`.

jianyangli:
@mkelley I rebased this branch to merge, but then the read the docs build failed.  I don't understand the error and can't find any problem with my local tests:

/home/docs/checkouts/readthedocs.org/user_builds/sbpy/envs/325/lib/python3.9/site-packages/sbpy/data/orbit.py:docstring of sbpy.data.orbit.Orbit.D_criterion:36: ERROR: Unexpected indentation.

Do you know what's the problem?  Thanks.

jianyangli:
Lines 815 - 817 are now identical to lines 757 - 759, but the latter passed with no problem.  I also removed the blank line following these lines, no luck either.

Since this is not a required test, should we just go ahead and merge, and diagnose the problem later?  I doubt it's the problem with the code.

mkelley:
@jianyangli  Got it!  I tested this out on a separate branch and found that the error is due to the indentation on lines 807 - 809.  Try reformatting the 'References' section to avoid this indentation.

jianyangli:
Thank you @mkelley !  It took me some commits to fix all the bugs and citations in the docstrings, but glad it finally worked.

mkelley:
Looks great!

PR #327, opened 2022-04-29
Update readthedocs config

mkelley:
Last build was failing.  It isn't clear to me exactly why, but may be due to graphviz missing.


jianyangli (approved):
Although I'm not really familiar with the setup of readthedocs, the change is small enough that I can probably be confident enough to say it looks good, as long as it passes the tests :)

PR #336, opened 2022-05-25
Vectorial model rewrite and update associated tests and documentation

sjoset:
This is a rewrite of the vectorial model to utilize numpy where possible, assemble results neatly, and bring the results better in line with the fortran version.

All tests in sbpy/activity/gas/tests/test_core.py passed and flake8 was clear.


sbpy/activity/gas/core.py
```diff
@@ -1046,314 +1147,315 @@ def _make_radial_logspace_grid(self):
...
+            num=self.grid.radial_points, endpoint=True
         )
 
-    def _compute_fragment_density(self):
-        """ Computes the density of fragments as a function of radius
+    def _outflow_axis_sampling(self,
+                               x: np.float64,
+                               y: np.float64,
+                               theta: np.float64
+                               ) -> tuple[np.ndarray, np.ndarray]:
```
mkelley, line 1162: I think for this to work, you'll need to use `Tuple` from the typing module, rather than the `tuple` object.

mkelley: This is what causes the tests to fail.

mkelley:
@sjoset There are some lines of code not being covered by the tests.  See the coverage report for details: https://app.codecov.io/gh/NASA-Planetary-Science/sbpy/pull/336/blob/sbpy/activity/gas/core.py  Most of it are trivial branches, but the `q_t` parameter and the `binned_production` method are especially important.  Never mind the scipy import tests and line 497 in the GasComa class, I'll get those separately.

mkelley:
That failing test will be addressed with #374

mkelley (changes requested):
Could you make an entry for this update in CHANGES.rst under New Features, including the improvements and the new `binned_production` constructor?  Also, add any API changes to the relevant section (e.g., `angular_substeps` was removed from `VectorialModel`).

Finally, remove the warning in the documentation (gas.rst).

Do you have or do you plan to have a citeable reference for the comparison to the FORTRAN version?

mkelley (approved):

mkelley:
Thanks for the fixes and updates!

PR #338, opened 2022-05-26
Fix remote_data move

mkelley:
The `remote_data` decorator cannot be imported from astropy as of 5.1.  Instead, use `@pytest.mark.remote_data`.


jianyangli (approved):
All looks good to me.  And I don't see any missed `@remote_data` instance.

mkelley:
Done!

PR #337, opened 2022-05-26
Documentation enhancements

mkelley:
I'm reviewing the documentation in preparation for a tutorial session at an upcoming meeting (LSST Solar System Science Collaboration Readiness Sprint).  First commit is something I've wanted for a while: to expose each of the data object documentation pages on the front page of the documentation.

PR #339, opened 2022-05-26
Fill delta and phase angle from rh and solar elongation

mkelley:
`Ephem` objects can be manually populated with arbitrary data.  Add a method that helps users generate geometrically sound values:

```python
>>> import astropy.units as u
>>> from sbpy.data import Ephem
>>> eph = Ephem.from_dict({'rh': [2] * u.au,
...                        'solar_elongation': [90] * u.deg})
>>> eph.fill_delta_and_phase()
>>> print(eph)
<QTable length=1>
   rh   solar_elongation       delta              phase       
   AU         deg                AU                deg        
float64     float64           float64            float64      
------- ---------------- ------------------ ------------------
    2.0             90.0 1.7320508075688772 30.000000000000004
```

### DIFF
--- a/docs/sbpy/data/ephem.rst
+++ b/docs/sbpy/data/ephem.rst
@@ -2,14 +2,62 @@
 Using Ephem
 ===========
 
-As shown above (:ref:`How to use Data Containers`),
-`~sbpy.data.Ephem` objects can be created on the fly. However,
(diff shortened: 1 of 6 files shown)


mkelley:
First needs PR #337

mkelley:
Maybe I could rework this into `fill(field)`, where field is `rh`, `delta`, `phase`, or `solar elongation`?

Issue #340, opened 2022-05-27
A/ designations do not parse


mkelley:
An A/ designation indicates an asteroidal object (in appearance) in a cometary orbit: https://minorplanetcenter.net/mpec/K18/K18H54.html

A/ objects do not parse as cometary or asteroidal:

```python
>>> from sbpy.data import Names
>>> Names.asteroid_or_comet('A/2018 V3')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/disks/data0/astro/Projects/sbpy/sbpy/data/names.py", line 597, in asteroid_or_comet
    raise TargetNameParseError('Target nature unclear.')
sbpy.data.names.TargetNameParseError: Target nature unclear.
```

**What did you expect?**
A/ should parse as an asteroid.

PR #343, opened 2022-06-04
Allow and test asteroids with A/ designations.

mkelley:
Addresses #340

### DIFF
--- a/sbpy/data/names.py
+++ b/sbpy/data/names.py
@@ -10,6 +10,7 @@ created on August 28, 2017
 
 """
 
+import re
 from ..exceptions import SbpyException
 
 __all__ = ['Names', 'TargetNameParseError', 'natural_sort_key']
@@ -47,7 +48,6 @@ def natural_sort_key(s):
     ['2P/Encke', '9P/Tempel 1', '10P/Tempel 2', '101P/Chernykh']
 
     """
-    import re
     keys = tuple()
     for k in re.split('([0-9]+)', str(s)):
         keys += (int(k) if k.isdigit() else k,)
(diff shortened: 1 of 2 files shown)


sbpy/data/names.py
```diff
@@ -342,7 +342,7 @@ def parse_comet(s):
                 if len(el[0]) > 0:
                     typnumber = el[0].replace('/', '')
                     try:
-                        r['type'] = re.findall('[PDCXAI]', typnumber)[0]
+                        r['type'] = re.findall('[PDCXI]', typnumber)[0]
```
jianyangli, line 345: Looks like the type "I" is included here, but not matched in the pattern in line 311.  Have we decided whether and how to include interstellar objects?  I think we should: 1) remove "I" here if it's not implemented to avoid confusion in the code, and 2) mention in the document that interstellar objects are not parsed at this time, but will be implemented later.

sbpy/data/names.py
```diff
@@ -475,14 +476,16 @@ def parse_asteroid(s):
...
+               '|^A/([12][0-9][0-9][0-9] [A-Z][0-9]+)'
+               # asteroids with cometary orbits [12]
                )
 
         # regex patterns that will be rejected
-        rej_pat = ('([1-2][0-9]{0,3}[ _][A-Z][0-9]*(\b|$))'
+        rej_pat = ('([CPXD]/[1-2][0-9]{0,3}[ _][A-Z][0-9]*(\b|$))'
                    # comet desig
-                   '|([1-9][0-9]*[PDCXAI]\b)'
+                   '|([1-9][0-9]*[PDCXI]\b)'
```
jianyangli, line 486: Again, same comment about "I" type.

jianyangli (commented):
Hi @mkelley , I just left a few comments and rebased this branch to bring it up to date with the main.  Hope that's fine with you.

sbpy/data/names.py
mkelley: Good catch.  I've removed the I here.  I have the documentation addressed in #337 : https://sbpy--337.org.readthedocs.build/en/337/sbpy/data/names.html

sbpy/data/names.py
mkelley: And here, since these patters are rejecting things that are not asteroidal, I think we want the I.  I've fixed line 484 and updated the comments.

jianyangli (approved):
Looks good to me!

Issue #342, opened 2022-06-03
Update Colina et al. solar spectrum URL


mkelley:
Now at: https://archive.stsci.edu/hlsps/reference-atlases/cdbs/current_calspec/

PR #345, opened 2022-06-06
Fix and test STScI Sun URLs

mkelley:
Address #342 and update weekly tests to hopefully more quickly find issues like this.  Also, fix a bug in the CI github workflow that was not including remote tests.


jianyangli (commented):
@mkelley I'm not really familiar with the test set up, so I don't feel qualified to review the .yml and tox.ini.  Maybe you should get someone else to check those parts?

mkelley:
Maybe @migueldvb can give them a quick look?

jianyangli (approved):

migueldvb:
Looks great, sorry I missed it.

Issue #335, opened 2022-05-16
Improve weekly test stability


mkelley:
Weekly doc tests fail when JPL SBDB / Horizons updates the absolute magnitude of Ceres, which occurs every month or so.  Being that Ceres is a well-observed object, I suspect it is not a true improvement on the absolute magnitude, but instead a reflection of the method and data sources.  There are several lines of code that need to be updated whenever this happens.

I propose that we find all doc tests that rely on `from_phys` or `from_horizons` for Ceres be skipped or updated to use hard coded values (the latter is preferred).

jianyangli:
OK, I can take care of this.

PR #352, opened 2022-06-24
Improve weekly test stability

jianyangli:
Address #335


jianyangli:
Looks like this PR should wait until after #338 is merged.  Some test failures due to the use of `@remote_data`, which will be fixed by #338

jianyangli:
@mkelley I only changed 5 files instead of 9 as reported by the PR because rebase included those you changed in PR #338 .  FYI for your review.  Thanks.

mkelley (approved):

PR #364, opened 2022-07-15
Fix `DataClass.__setitem__`

jianyangli:
This fix allows `DataClass.__setitem__` to set an existing column with an alternative name, rather than adding a new column.


jianyangli:
I also improved `_translate_columns` and `__contains__`.  The change and rationale are in the commit comments.  Please let me know if those make sense or perhaps overengineered.  Thanks.

mkelley (approved):
Thanks, these changes are great.  Just a couple minor comments.  I'll approve now and you can address them as you see fit.

Issue #203, opened 2019-07-23
Combining DataClass objects


mkelley:
This is a request for 
* [x] a new feature
* [ ] an enhancement to existing sbpy functionality
* [ ] somethings else: [explain here]

The requested changes will be implemented by
* [ ] me
* [x] the sbpy developers 

**High-level concept**
Provide a mechanism to combine `DataClass` objects that have different fields.

**Explain the relevance to sbpy**
`DataClass` objects have flexibility beyond what `QTable` is designed for.

**Proposal details**
Something like the following behavior would be useful:
```python
from astropy.time import Time
from sbpy.data import Orbit

orbits = Orbit.from_horizons(
        '46P', 'designation', epochs=(obs_dates['spitzer'].jd.mean(), obs_dates['bass']),
        closest_apparition=True)
for planet in (399, 599):
    orbits.add_row(Orbit.from_horizons(planet, 'majorbody', epochs=obs_dates['bass'])[0])
```

However, `DataClass.add_row` does not exist.  `DataClass.table.add_row` does exist, but:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-15-02e3f7890044> in <module>()
      3         closest_apparition=True)
      4 for planet in (399, 599):
----> 5     orbits.table.add_row(Orbit.from_horizons(planet, 'majorbody', epochs=obs_dates['bass'])[0])

/home/msk/local/lib/python3.6/site-packages/astropy-3.2.1-py3.6-linux-x86_64.egg/astropy/table/table.py in add_row(self, vals, mask)
   2399              3   6   9
   2400         """
-> 2401         self.insert_row(len(self), vals, mask)
   2402 
   2403     def insert_row(self, index, vals=None, mask=None):

/home/msk/local/lib/python3.6/site-packages/astropy-3.2.1-py3.6-linux-x86_64.egg/astropy/table/table.py in insert_row(self, index, vals, mask)
   2488 
   2489             if len(self.columns) != len(vals):
-> 2490                 raise ValueError('Mismatch between number of vals and columns')
   2491 
   2492             if mask is not None:

ValueError: Mismatch between number of vals and columns
```

mommermi:
This is an interesting and non-trivial problem. Orbits for different objects use different parameters:

```
from sbpy.data import Orbit

orbits = Orbit.from_horizons(
    '46P', 'designation',
    closest_apparition=True)
print(orbits.field_names)

[...]

PR #367, opened 2022-09-21
Enable combining `DataClass` objects

jianyangli:
This PR responds to issue #203 to provide a mechanism to combine `DataClass` objects.  New methods implemented include:

- `.add_row()`
- `.add_column()`
- `.join()`

These methods allow for the same flexibilities provided by other `DataClass` methods, such as column name alias, expanding table with a new `DataClass` or `astropy.table.Table` object, new row, or new column with different shapes from the original table.

The implementation tries to make use of the existing functionalities from `astropy.table.Table` and `DataClass` as much as possible.  `join` function uses `astropy.table.vstack`.  `add_row` uses `DataClass.from_rows` and `join`.  `add_column` uses `DataClass.apply`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,7 +22,7 @@ sbpy.data
 - Added ``DataClass.__contains__`` to enable `in` operator for ``DataClass``
   objects. [#357]
 
-- Added ``DataClass.add_row``, ``DataClass.add_column``, ``DataClass.join``
+- Added ``DataClass.add_row``, ``DataClass.vstack``
   methods. [#367]
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,7 +23,7 @@ sbpy.data
   objects. [#357]
 
 - Added ``DataClass.add_row``, ``DataClass.add_column``, ``DataClass.join``
-  methods. [#xxx]
+  methods. [#367]
 
 
 Bug Fixes
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,9 @@ sbpy.data
 - Added ``DataClass.__contains__`` to enable `in` operator for ``DataClass``
   objects. [#357]
 
+- Added ``DataClass.add_row``, ``DataClass.add_column``, ``DataClass.join``
+  methods. [#xxx]
+
 
 Bug Fixes
 ---------
--- a/docs/sbpy/data/dataclass.rst
+++ b/docs/sbpy/data/dataclass.rst
@@ -411,21 +411,17 @@ directly addressing them:
     >>> obs['ra']
     <Quantity [10.323423, 10.333453, 10.343452] deg>
 
-More complex data table modifications are possible by directly
-accessing the underlying `~astropy.table.QTable` object as shown below.
-
-`~sbpy.data.DataClass` provides a direct interface to the table
-modification functions provided by `~astropy.table.Table`:
-`~astropy.table.Table.add_row`, `~astropy.table.Table.add_column`,
-`~astropy.table.Table.add_columns`, etc. For instance, it is trivial to add
-additional rows and columns to these objects.
+The basic functionalities to modify the data table are implemented in
+`~sbpy.data.DataClass`, including adding rows and columns and join a
+DataClass with another DataClass object or an `~astropy.table.Table`
+object.
 
 Let's assume you want to add some more observations to your ``obs``
 object:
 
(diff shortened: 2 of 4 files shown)


docs/sbpy/data/dataclass.rst
```diff
@@ -477,16 +472,42 @@ Similarly, existing columns can be modified using:
...
+    10.333453 -12.41562  2451523.7345      i       V     ———
+    10.343452 -12.40435  2451523.8525      R       R     ———
+     10.25546  -12.3946 2451523.94653      V       i     ———
+    20.223423  12.42123  2451623.6234     --      --    10.1
+    20.233453  12.41562  2451623.7345     --      --    12.3
+    20.243452  12.40435  2451623.8525     --      --    15.6
+
+Note that the data table to be joined doesn't have to have the same
+columns as the original data table.  The empty field will be automatically
+masked out.
```
mkelley, line 503: Can we have this as an option that can be turned off?  Or maybe it is off by default?  As a programmer, I can see some value in having an exception raised when the columns do not match.

sbpy/data/core.py
```diff
@@ -934,3 +938,135 @@ def verify_fields(self, field=None):
...
+        col : `~astropy.table.Column` object, or sequence
+            The column to be added.  Must have the same length as the
+            existing table.
+        name : array-like str, optional
+            Specify the name of added column.  If not ``None``, then
+            takes precedence over ``col.name``.
+        unit : array-like `~astropy.units` objects or str, optional
+            Unit to be applied to the new column.  If not ``None``, then
+            take precedence over ``col.unit``.
+        """
```
mkelley, line 1022: Missing an example.

sbpy/data/core.py
```diff
@@ -934,3 +938,135 @@ def verify_fields(self, field=None):
...
+        if name is None:
+            name = getattr(col, 'name', '')
+        if unit is None:
+            unit = getattr(col, 'unit', None)
+        if name in self:
+            raise DataClassError('Column {} already exists.'.format(name))
+        self.apply(col, name=name, unit=unit)
+
+    def join(self, data):
+        """Join another DataClass object to the end of DataClass
```
mkelley, line 1032: I think this is vertical stacking, i.e., `vstack`.  As I understand it, table joining is a different operation, borrowed from SQL, and according to wikipedia, borrowed from relational algebra: https://en.wikipedia.org/wiki/Join_(SQL)  It is a very powerful way to combine tables, so I'd rather rename this as `vstack` and reserve `join` for a future implementation of `astropy.table.join`.

mkelley (commented):
Another belated review, but thanks for the PR!  I do have two major comments on join.

sbpy/data/core.py
jianyangli: I'm going to remove `.add_column()`.  I forgot what my original rationale was to add this method, maybe just to maintain a similar class interface as `astropy.table.Table`.  But now I don't think this is necessary if we have `.apply()` to implement the similar functionality.

docs/sbpy/data/dataclass.rst
jianyangli: Since this method is basically equivalent to `astropy.table.Table.vstack()` (as you correctly point out), I'll just use the 'join_type' keyword parameter to deal with the different cases.

`vstack` has default join_type as 'outer', which means merge with all columns and mask out the cells with no values.  I think it's a good idea to maintain the consistency, and just let `Table.vstack()` raise any necessary exceptions.

Sounds good to you?

mkelley: Ah, yes, that sounds good.

jianyangli:
@mkelley OK this one is ready for review, too.  Thanks.

CHANGES.rst
```diff
@@ -22,6 +22,9 @@ sbpy.data
 - Added ``DataClass.__contains__`` to enable `in` operator for ``DataClass``
   objects. [#357]
 
+- Added ``DataClass.add_row``, ``DataClass.add_column``, ``DataClass.join``
```
mkelley, line 25: Correct with actual methods added.

jianyangli: oops, thanks for catching that!

mkelley (approved):

mkelley:
That failing test is something that should be marked as a remote test, and I'll have a fix in #374

jianyangli:
OK, should I go ahead with merge, or wait for the fix first?

PR #371, opened 2023-05-02
Add CALSPEC model solar spectrum

mkelley:
Adding support for a R=5000 model solar spectrum from STScI's CALSPEC database.


mkelley:
Requires #370 .

jianyangli:
I see some conflicts/duplicates in this PR with #370 .  I assume this PR will be rebased and duplicates resolved before we should review and merge?

mkelley:
The decrease in coverage is due to adding a remote test.  So, this is ready for review.

jianyangli (approved):
Just curious: In your recent few PRs, you are changing single quote (') to double quote (").  Is that something new in Python 3.10 or 3.11?  I'm still using 3.8, but wondering whether I need to do the same to my personal code if I upgrade.  Thanks.

mkelley:
It isn't important, just a style preference.  I like to use the Black
formatter and it automatically does the change "The main reason to
standardize on a single form of quotes is aesthetics. Having one kind of
quotes everywhere reduces reader distraction."
https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings

PR #373, opened 2023-06-07
EXP: Add tomli to doc build req

pllim:
Just trying to narrow down what broke the docs as reported in https://github.com/astropy/sphinx-astropy/issues/61

😬

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -48,7 +48,7 @@ test =
     pytest-xdist
     coverage
 docs =
-    tomli; python_version<3.11
+    tomli; python_version<'3.11'
     sphinx-astropy>=1.3
     pytest
     matplotlib>=3.0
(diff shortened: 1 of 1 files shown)


pllim:
@mkelley , can you pls approve the Actions? 🙏

mkelley (approved):

pllim:
Ugh nvm... I think I found the reason and the fix is indeed upstream in sphinx-astropy. For whatever reason, `pip` was choking on the requirement name "v2". Maybe it is some obscure bug they finally fixed in Python 3.11. I will have to make another release upstream.

mkelley:
Got it, thanks for looking into it!

pllim:
Thanks for reporting this! It was an oversight on my part.

Issue #110, opened 2019-02-17
Afρ to dust cross section


mkelley:
Add methods to convert Afρ / εfρ to and from dust cross section.

PR #376, opened 2023-06-12
New coma continuum models and convert Afrho/Efrho to/from cross-sectional area.

mkelley:
Draft coma continuum models: reddened Afρ model (scattered light) + εfρ model (thermal emission).

Also, add methods to `Afrho` and `Efrho` to convert to and from cross-sectional area.


mkelley:
Closes #110

Issue #377, opened 2023-06-30
Website linked to 'latest' documentation


mkelley:
The website is currently linked to the latest documentation.  What I've been noticing is that most users are not installing the latest version, but whatever is available via pip or conda, i.e., the last stable version.  Should we update the sbpy.org link to point to the stable version?  Or, just to sbpy.readthedocs.io and we can configure latest vs. stable on the RTD end (currently it is stable).

jianyangli:
I think it's a good idea to link to the stable version.  Advanced users will find out the latest dev version by themselves if they need.  Most users don't care about the dev version anyway.  Linking to readthedocs website is also fine.  Would the readthedocs website be synced with the version available via pip or conda?

jianyangli:
Why do I see version number [v0.1.dev](https://sbpy.readthedocs.io/en/latest/index.html#) for the latest version?  Shouldn't it be v0.4.dev?

Also, the current version on sbpy.org also needs to be updated to 0.4.

mkelley:
I'm trying to understand that, too.  It might be https://docs.readthedocs.io/en/stable/build-customization.html?highlight=setuptools_scm#avoid-having-a-dirty-git-index which is an issue that astropy also had.

mkelley:
Readthedocs builds stable versions of the documentation whenever a release is tagged in githb.  v0.4 is available there now.

mkelley:
These issues seem to be fixed now: #378

PR #378, opened 2023-06-30
Website: update online doc links, CI badge, sbpy version

mkelley:

PR #379, opened 2023-06-30
Avoid a dirty git index with readthedocs

mkelley:
Following discussion at #377 , this PR attempts to fix the version in RTD's "latest" documentation: https://sbpy.readthedocs.io/en/latest

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,6 +63,7 @@ distribute-*.tar.gz
 *py#
 .firebase
 .tmp
+coverage.xml
 
 # Mac OSX
 .DS_Store
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -7,9 +7,12 @@ version: 2
 build:
   os: ubuntu-20.04
   tools:
-    python: "3.9"
+    python: "3.10"
   apt_packages:
     - graphviz
+  jobs:
+    pre_install:
+      - git update-index --assume-unchanged docs/conf.py
 
 sphinx:
   builder: html


jianyangli (commented):
Mike, I have to admit that I'm not familiar with the readthedocs build configuration here.  I know the "jobs" block tells git to ignore the changes in conf.py, but don't know what problems it's going to solve.  Can you tell me a little more or point me to the resources to help me understand?  Thanks.

mkelley:
The problem is described at: https://docs.readthedocs.io/en/stable/build-customization.html?highlight=setuptools_scm#avoid-having-a-dirty-git-index

We are using setuptools_scm, which derives the version number from git tags and commits.  For every commit beyond the last tag it appends a dev string and a number: https://github.com/pypa/setuptools_scm/

When any git-tracked file is changed, whether or not it is committed, the version is incremented.  Apparently readthedocs needs to change docs/conf.py, so this edit tells git to consider that unchanged, which should keep setuptools_scm from incrementing the version, and keeping it clean, like sbpy v0.1.dev172 instead of sbpy v0.1.dev172+g39b0c3c .

Unfortunately, I don't see a way to review the outcome of this edit.

Here is the related astropy PR https://github.com/astropy/astropy/pull/14860

mkelley:
The build on #378 (i.e., without this edit) was already using the right tag, incrementing beyond v0.4.0 https://sbpy--378.org.readthedocs.build/en/378/  So, I have no idea why "latest" doesn't use it.  If this doesn't work, I'll contact RTD to see if they understand it.

jianyangli:
I see.  Now it makes sense to me.  Thanks.

I was about to ask you the question of the version number.  So let's just go with it now and see how it goes.

jianyangli (approved):

mkelley:
Thanks 🤞🏻

mkelley:
https://sbpy.readthedocs.io/en/latest/#  Closer!  But not quite there.

mkelley:
I've reviewed what setuptools_scm does [1], and I think this is as good as it will get.  

[1] https://github.com/pypa/setuptools_scm/tree/main#default-versioning-scheme

PR #380, opened 2023-07-17
WIP: Basic thermal models

jianyangli:
This is an implementation of the basic thermal models: STM, FRM, and NEATM.  It's still a WIP.  I'd like to start a PR here for input on everything, API, class inheritance structure, as well as the implementation.

### DIFF
--- a/sbpy/thermal/core.py
+++ b/sbpy/thermal/core.py
@@ -25,27 +25,41 @@ class ThermalClass(abc.ABC):
     such as integration of total flux based on a temperature distribution.
     """
 
-    @u.quantity_input(rh=u.km, R=u.km)
-    def __init__(self, rh, R, albedo=0.1, emissivity=1., beaming=1.):
+    @quantity_to_dataclass(rh=(Ephem, 'rh'))
+    @dataclass_input(phys=Phys)
+    def __init__(self, rh, phys):
         """
         Parameters
         ----------
-        rh : u.Quantity
-            Heliocentric distance
-        R : u.Quantity
-            Radius of asteroid
-        albedo : float, u.Quantity
-            Bolometric Bond albedo
-        emissivity : float, u.Quantity
-            Emissivity of surface
-        beaming : float, u.Quantity
-            Beaming parameter
+        rh : `~sbpy.data.Ephem`, dict_like, number, or
+            `~astropy.units.Quantity`
+            If `~sbpy.data.Ephem` or dict_like, ephemerides of the object that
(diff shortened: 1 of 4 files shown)


sbpy/thermal/core.py
```diff
@@ -5,66 +5,274 @@
...
-        >>> from sbpy.data import Ephem, Phys
-        >>> epoch = Time('2019-03-12 12:30:00', scale='utc')
-        >>> eph = Ephem.from_horizons('2015 HW', location='568', epochs=epoch) # doctest: +REMOTE_DATA
-        >>> phys = PhysProp('diam'=0.3*u.km, 'pv'=0.3) # doctest: +SKIP
-        >>> lam = np.arange(1, 20, 5)*u.micron # doctest: +SKIP
-        >>> flux = STM.flux(phys, eph, lam) # doctest: +SKIP
-
-        not yet implemented
+        lon : float
+            Longitude in radiance
```
mkelley, line 87: What does this mean, longitude/latitude in "radiance"?

sbpy/thermal/core.py
```diff
@@ -5,66 +5,274 @@
...
+            else:
+                m = np.array([[0, 0, 1], [0, 1, 0], [-1, 0, 0]])
+        else:
+            m = twovec([sublon.to_value('deg'), sublat.to_value('deg')], 0,
+                       [0, 90], 2).T
+        return m
+
+    @u.quantity_input(wave_freq=u.m, delta=u.m, lon=u.deg, lat=u.deg,
+                      equivalencies=u.spectral())
+    def fluxd(self, wave_freq, delta, sublon, sublat, unit='W m-2 um-1',
```
mkelley, line 141: I like the idea of using sublon and sublat, but can we also have a way to use obliquity and phase angle?

sbpy/thermal/core.py
jianyangli: Ah, typo.  Should be radian.

sbpy/thermal/core.py
jianyangli: This method calculates the total flux density by integrating over the whole visible part of the surface of an object, provided that the temperature distribution is available by `.T(lon, lat)`.  So as long as a mechanism is provided to convert (obliquity, phase) or any other combination to (sublon, sublat) in a subclass or inherited `.fluxd` method, we can use it in any circumstance.  I think this will also work for an irregular shape model, but haven't thought about it thoroughly or tested it yet.

Issue #389, opened 2023-11-10
SpectralGradient.from_color is inverted for linear inputs


mkelley:
**High-level problem description**
The color is wrong for linear inputs:

```python
>>> wavelengths = [1, 2] * u.um
>>> red_color = 2
>>> SpectralGradient.from_color([1, 2] * u.um, red_color)
<SpectralGradient -6.66666667 % / 100 nm>
>>> blue_color = 0.5
>>> SpectralGradient.from_color([1, 2] * u.um, blue_color)
<SpectralGradient 6.66666667 % / 100 nm>
```

The colors are inverted.   There are two problems:

1) The wavelengths are reversed on line 201:  `dw = lambda_eff[0] - lambda_eff[1]`, the convention for the array is [blue, red], so this should instead by [1] - [0]

2) Fixing that will cause a problem for magnitude inputs.  The current behavior is correct:

```python
>>> SpectralGradient.from_color([1, 2] * u.um, 1 * u.mag)
<SpectralGradient 8.61011004 % / 100 nm>
```

The convention for color in magnitudes is blue - red, so 1 mag is a red color, and results in a red spectral gradient.  However, if (1) is fixed, the code will return a blue gradient for a red color index.  This must also be addressed.

PR #392, opened 2023-12-06
Provide correct slope for color ratios with `SpectralGradient.from_color`

mkelley:
Addresses #389

### DIFF
--- a/sbpy/spectroscopy/core.py
+++ b/sbpy/spectroscopy/core.py
@@ -9,7 +9,7 @@ import numpy as np
 import astropy.units as u
 
 
-__all__ = ['Spectrum', 'SpectralModel', 'SpectralGradient']
+__all__ = ["Spectrum", "SpectralModel", "SpectralGradient"]
 
 __doctest_requires__ = {
     "SpectralGradient": ["synphot"],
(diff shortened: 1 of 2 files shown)


hhsieh00:
Sorry, just getting to this now.  In the code diff view, Codecov indicates that there are several new lines not covered by tests.  Is this a problem (I note that the Codecov summary report on this page says that all "coverable" lines are covered)?  Otherwise, the changes look good to me.

mkelley:
Those reported lines don't make sense.  They show up as covered in the codecov app: https://app.codecov.io/gh/NASA-Planetary-Science/sbpy/pull/392/blob/sbpy/spectroscopy/core.py

mkelley:
I've rebased.  Maybe that will fix them?  If not, I think the coverage report should take precedence.

hhsieh00 (approved):
Okay, that sounds fine to me.  Yeah, the contradiction between the different reports is a bit strange (I still see the comments in the code diffs).

mkelley:
I can refresh the codecov check and it'll report 100% covered as it should.  But the diff report always lists those lines as uncovered.  Anyway, something to keep in mind in other reviews.

PR #393, opened 2024-01-18
Initialize CircularAperture from other apertures

mkelley:
The `Afrho` and `Efrho` classes can take a `Quantity` or an `Aperture` as an input for the photometric aperture radius, and they will convert it to a coma equivalent radius, as needed.  While writing new code for PR #376, I needed to repeat that same Quantity/Aperture conversion code, and also repeat tests for each case.  This seemed like an opportunity for an enhancement that could simplify the code.

Each of the `Aperture` classes can be converted to an effective circular aperture based on the assumption of a 1/rho coma.  This PR enables any of the `Aperture` classes to directly initialize a `CircularAperture` object using that coma equivalent aperture conversion:
```python
>>> import astropy.units as u
>>> from sbpy.activity import CircularAperture, RectangularAperture
>>>
>>> rect = RectangularAperture([1, 5] * u.arcsec)
>>> circ = CircularAperture.from_coma_equivalent(rect)
>>> print(circ)
Circular aperture, radius 1.0522971172731228 arcsec
```

Or, a radius may be given as a `Quantity`, in which case the "coma equivalent" radius is the same value:
```python
>>> circ = CircularAperture.from_coma_equivalent(10 * u.arcsec)
>>> print(circ)
Circular aperture, radius 10.0 arcsec
```

With this change, the `Afrho`, `Efrho`, and any similar code that needs the effective aperture radius for a 1/rho coma can pass their input parameters to `CircularAperture` without the need to test for `Quantity` vs. `Aperture` input, and the developers do not need to be concerned with the conversions for testing coverage:

```python
# works for aper as Quantity or Aperture:
rho = CircularAperture.from_coma_equivalent(aper).dim
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,15 +13,19 @@
 New Features
 ------------
 
+sbpy.activity
+^^^^^^^^^^^^^
+- New `sbpy.activity.CircularAperture.from_coma_equivalent()` to immediately
+  create a `CircularAperture` from any other `Aperture` given a nominal coma
+  surface brightness distribution. [#393]
+
 sbpy.utils
 ^^^^^^^^^^
-
 - New `required_packages` and `optional_packages` functions to test for the
   presence of required and optional packages.
 
 sbpy.utils.decorators
 ^^^^^^^^^^^^^^^^^^^^^
-
 - New `requires` and  `optionally_uses` function decorators to simplify testing
   for required and optional packages.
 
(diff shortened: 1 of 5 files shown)


hhsieh00 (approved):
All seems to check out.  Tested documentation examples on OSX and works fine.

mkelley:
Thanks!

Issue #19, opened 2017-12-04
Implement syndynes and synchrones


mkelley:

mkelley:
Pending `Ephemeris` class definition.

mkelley:
I'm considering class designs for implementing this feature.  At the heart, a syndyne generator is an integrator that connects the dots between particles of a single beta.  It needs the state of the target at the time of the observation, the list of betas and the ages to simulate, and an integrator.  In sbpy, we have an orbit and ephemeris class that can potentially be used to provide the state vector.

Thinking about API options.  Here is a first try.
```python
state = State(r, v, t)  # initialize state vector from position, velocity, and time of observation
state = State.from_ephem(eph)  # initialize from Ephem object (requires x, y, z, vx, vy, vz, and time fields)
state = State.from_orbit(orbit)  # initialize from Orbit object

syn = Syndynes(state)  # using defaults
syn = Syndynes(state, betas=betas, ages=ages)  # specific betas and ages
syn = Syndynes(state, integrator='quad')  # specify integrator or integration method to use (if we have options)

syn.betas[0]  # the first beta quantity
syn.syndyne[0]  # get state vectors for all particles with beta=betas[0]
syn.ages[0]  # the first age
syn.synchrone[0]  # get state vectors for all particles with age=ages[0]

syn.plot_syndynes()  # default observer is the Earth
syn.plot_synchrones()

# how to observe at other locations?
syn.plot_syndynes(observer=r)  # specify observer's location with a heliocentric position vector?
syn.plot_syndynes(observer=state)  # specify observer's location with a State object?
syn.plot_syndynes(observer=coords)  # specify observer's location with SkyCoords object?
```

I'm going to think about whether or not an `astropy.coordinates.SkyCoord` object could be used in lieu of `State`.  That might help us take advantage of the astropy reference frames.

jianyangli:
I like the top-level design.  It includes everything I will need, at least, in a simple and clear way.

I see the need of using `SkyCoord` in many places in the implementation of plot-related methods, but I doubt it can be used in lieu of `State`, because you would need at least three coordinates together with times to uniquely solve the state.  Solving for state is not at the core of `Syndynes` class.

I suggest making the time parameter optional in the initiation of a `State` object with position and velocity, so the class supports relative time in the calculation of syndynes/synchrones.  The absolute time isn't important.  After all, ages will be relative.

[...]

PR #394, opened 2024-01-26
Add support for dynamical models and dust syndynes/synchrones.

mkelley:
Adds the capability to generate syndynes and synchrones.  This PR is essentially four enhancements: 

1. Add `State` objects to hold particle position, velocity, and time.
2. Add SPICE-equivalent "ephemeris time" capability to the `astropy.time` system.
3. Add the `DynamicalModel` abstract base class to define an API for solving the equations of motion, and implement a few simple models.
4. Add `SynGenerator` and supporting classes which can generate syndynes/synchrones.

Design choices are given below.

# `State` objects

Dynamical state is 3D position, velocity, and time.  The `sbpy.dynamics.State` object encapsulates these quantities using the `astropy.coordinates` sub-module.  State vectors are internally stored as a coordinate frame object with position and velocity data.  Example coordinate frame objects are `HeliocentricEclipticIAU76`, `FK5`, and `ICRS`.  Some coordinate frames require observation time for transformation to other coordinate frames, e.g., from a heliocentric to a barycentric frame.  Time is added to the frame as needed.  In addition, `sbpy` adds an `ArbitraryFrame`, which is used by default and cannot be converted to other frames.

The state's time attribute is typically an `astropy.time.Time` object.  However, when the `ArbitraryFrame` is used the state's time attribute may be an `astropy.units.Quantity`.

Create a `State` object for a particle at *x*=2 au, moving with *v_y*=30 km/s, on 2023 Dec 08:

```python
>>> from astropy.time import Time
>>> import astropy.units as u
>>> from sbpy.dynamics import State
>>>
>>> r = [2, 0, 0] * u.au
>>> v = [0, 30, 0] * u.km / u.s
>>> t = Time("2023-12-08")
>>> comet = State(r, v, t)
>>> comet
<State (<ArbitraryFrame Frame>):
 r
  [2. 0. 0.] AU
 v
  [ 0. 30.  0.] km / s
 t
  2023-12-08 00:00:00.000>
```

States can also be arrays of states:
```python
>>> len(comet)
1
>>> comets = State([r, r], [v, v], [t, t])
>>> len(comets)
2
>>> comet.r.shape  # [x, y, z]
(3,)
>>> comets.r.shape  # [[x, y, z], [x, y, z]]
(2, 3)
```

To specify a coordinate frame, use the `frame` keyword argument:

```python
>>> State(r, v, t, frame="heliocentriceclipticiau76")
<State (<HeliocentricEclipticIAU76 Frame (obstime=2023-12-08 00:00:00.000)>):
 r
  [2. 0. 0.] AU
 v
  [ 0. 30.  0.] km / s
 t
  2023-12-08 00:00:00.000>
```

## Alternative implementations

Initially, `State` stored its data as `r`, `v`, and `t` arrays.  This involved extra code for parameter checking and coordinate transformations.  Switching to using `BaseCoordinateFrame` objects simplified both of these aspects.  However, `t` still must be saved internally, as some coordinate frames do not use time.

`State` could instead use `astropy.coordinates.SkyCoord`.  `SkyCoord` can store 3D vectors, but is also designed to accommodate 2D representations and operations on the celestial sphere.  Dynamical state must only be 3D.  `BaseCoordinateFrame` data could also be 2D, but is more general and lacks many of `SkyCoord`'s 2D-based convenience methods.  I also found confusing `SkyCoord`'s tendency to duplicate position and velocity vectors in its coordinate frame objects:

```python
>>> coords = SkyCoord(ra=1 * u.deg, dec=2 * u.deg, distance=1 * u.au)
>>> coords
<SkyCoord (ICRS): (ra, dec, distance) in (deg, deg, AU)
    (1., 2., 1.)>

>>> coords.frame
<ICRS Coordinate: (ra, dec, distance) in (deg, deg, AU)
    (1., 2., 1.)>
```

Sub-classing `SkyCoord` or `BaseCoordinateFrame` would complicate the `State` API, especially if `SkyCoord` was sub-classed, whereas the goal is to keep the API simple.

`State` could have been based on  `Ephem`.  `Ephem` lacks coordinate frame information.  Functionally, this would be similar to saving `r`, `v`, and `t` as their own arrays.

# Ephemeris time

NAIF SPICE uses TDB seconds from the J2000 epoch for time coordinates.  This is not explicitly supported in `astropy.time`, so a new `SpiceEphemerisTime` has been created.  By importing the `sbpy.time` sub-module, it becomes available for use as the format "et":

```python
>>> from astropy.time import Time
>>> import sbpy.time
>>>
>>> j2000 = Time(0, format="et")
>>> j2000.iso
'2000-01-01 12:00:00.000'
>>> Time(759307746.954761, format="et")
<Time object: scale='tdb' format='et' value=759307746.954761>
```

Dynamical integrations (below) will convert `Time` objects to ephemeris time.

# Dynamical models

A lightweight approach to integrating the equations of motion has been implemented in `sbpy.dynamics`.  The new `DynamicalModel` abstract base class defines the API and implements solving the equations with `scipy.integrate.solve_ivp`.  The implementation is straightforward if we bump `sbpy`'s minimum supported `scipy` version from 1.3 to 1.4, which allows for arguments to be passed from `solve_ivp` to the integrated functions.

`solve_ivp` keyword arguments, such as solution tolerances, or the choice of integrator, are set during `DyanmicalModel` initialization.  The default arguments were designed for high precision: a relative tolerance of 2.3e-14, use of a function for the Jacobian matrix (as opposed to numerical estimates), and use of the [LSODA](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.LSODA.html) integrator.  All of these may be overridden by the user at the time of initialization.

Inheriting classes provide methods that calculate the derivatives of position and velocity, and the Jacobian matrix.  Time is given as a float in units of seconds, and position and velocity as a 6-element array (`[x, y, z, v_x, v_y, v_z]`) in units of kilometers and kilometers per second.

```python
@abc.abstractclassmethod
def dx_dt(cls, t: float, rv: np.ndarray, *args) -> np.ndarray:
    """Derivative of position and velocity.
    ...
    """

@abc.abstractclassmethod
def df_drv(cls, t: float, rv: np.ndarray, *args) -> np.ndarray:
    """Jacobian matrix, :math:`df/drv`.
    ...
    """
```

The solution is generated using the `solve` method, whose arguments are the initial dynamical state and the time at which the solution is desired.  Additional arguments for `dx_dt` and `df_drv` are possible, and necessary in order to pass the `beta` parameter for integrations considering radiation pressure.

Three dynamical model implementations are defined, `FreeExpansion`, `SolarGravity`, and `SolarGravityAndRadiationPressure`.  The following example integrates a pure-gravitational orbit around the Sun:

```python
>>> import astropy.units as u
>>> from sbpy.dynamics import State, SolarGravity
>>> state = State([1, 0, 0] * u.au, [0, 30, 0] * u.km / u.s, 0 * u.s)
>>> integrator = SolarGravity()
>>> t_final = 1 * u.year
>>> integrator.solve(state, t_final)
<State (<ArbitraryFrame Frame>):
  r
    [ 1.48146925e+08 -2.09358771e+07  0.00000000e+00] km
  v
    [ 4.137801   29.70907176  0.        ] km / s
  t
    1.0 yr>
```

# Syndynes and synchrones

The capability to produce syndynes and synchrones has been added to `sbpy.dynamics` with the `SynGenerator` class.  A `State` object is used for the dust source, and dust particle orbits are parameterized using *beta*, the ratio of the force from solar radiation to the force from solar gravity.

The source object state and dust particle states are integrated using the `DynamicalModel` framework.  The `SolarGravityAndRadiationPressure` model is used by default, but others may be provided at the time the `SynGenerator` object is initialized.  This framework should allow for more complex problems, such as planetary perturbations, or particle fragmentation, provided a dynamical model can be developed to support them.

```python
>>> r = [2, 0, 0] * u.au
>>> v = [0, 30, 0] * u.km / u.s
>>> t = Time("2023-12-08")
>>> comet = State(r, v, t)
>>> betas = [1, 0.1, 0.01, 0]
>>> ages = np.linspace(0, 100, 26) * u.day
>>> dust = SynGenerator(comet, betas, ages)
```

The integrated particle states are saved in the `particles` attribute.  However, it is more convenient get the syndynes and synchrones packaged as `Syndyne` and `Synchrone` objects, which are specialized `State` objects:

```python
>>> syndyne = dust.syndyne(0)  # get the first syndyne
>>> len(syndyne)
26
>>> syndyne.x
<Quantity [2.99195741e+08, 2.99284224e+08, 2.99549029e+08, 2.99988249e+08,
           3.00598735e+08, 3.01376158e+08, 3.02315083e+08, 3.03409054e+08,
           3.04650704e+08, 3.06031870e+08, 3.07543713e+08, 3.09176847e+08,
           3.10921470e+08, 3.12767487e+08, 3.14704634e+08, 3.16722591e+08,
           3.18811090e+08, 3.20960013e+08, 3.23159471e+08, 3.25399886e+08,
           3.27672045e+08, 3.29967158e+08, 3.32276893e+08, 3.34593405e+08,
           3.36909361e+08, 3.39217944e+08] km>
```

If an observer is provided to the `SynGenerator` object at initialization, then the syndyne/synchrone object will have a `coords` attribute, which is a `SkyCoord` object containing the observed coordinates of the syndyne/synchrone test particles.

```python
>>> observer = State([0, 1, 0] * u.km, [30, 0, 0] * u.km / u.s, comet.t)
>>> dust = SynGenerator(comet, betas, ages, observer=observer)
>>> syndyne = dust.syndyne(0)
>>> syndyne.coords # equivalent to: observer.observe(syndyne)
<SkyCoord (ArbitraryFrame): (lon, lat, distance) in (deg, deg, km)
    [(359.99999981, 0., 2.99195741e+08),
     (359.99960848, 0., 2.99284224e+08),
     (359.99687763, 0., 2.99549030e+08),
     (359.98950966, 0., 2.99988254e+08),
     (359.97528925, 0., 3.00598763e+08),
     (359.95212067, 0., 3.01376263e+08),
     (359.91806084, 0., 3.02315392e+08),
     (359.87134702, 0., 3.03409819e+08),
...
```

Syndynes and synchrones may be converted to `Ephem` objects.  The sbpy `DataClass` fields were updated to include the beta parameter (beta_rad) and relative time (t_relative).

## Alternative implementations

Initially, `SynGenerator` had a two step process:

```python
>>> dust = SynGenerator(comet, betas, ages)
>>> dust.solve()
```

The initial states were generated at initialization, and final states by the `solve()` method.  Two steps allows for the initial particle states to be modified by the user before being integrated, e.g., to give them a non-zero ejection velocity.  This is still possible in the implemented design, but would be instead addressed by sub-classing and overriding `initialize_states`:

```python
>>> class AlternativeSynGenerator(SynGenerator):
...     def initialize_states(self):
...         super().initialize_states()
...         # then modify self.initial_states as needed
```

# Other changes

There are some improvements to the activity.dust documentation.  Some package infrastructure files have been edited with small improvements.  The pytest configuration now respects the options in setup.cfg.

Addresses #19.


mkelley:
@hhsieh00 FYI, readthedocs builds documentation on pull requests (see the "checks" above).  For your convenience, the time and dynamics documentation are at:

https://sbpy--394.org.readthedocs.build/en/394/sbpy/time.html
https://sbpy--394.org.readthedocs.build/en/394/sbpy/dynamics.html

mkelley:
Unfortunately, removing the blackslashes causes Sphinx to interpret those lines as parts of an enumerated list, so "(2001) Einstein" becomes "2001. Einstein".  However, we can fix that by following the closing parenthesis with a non-breaking space.

https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#enumerated-lists

mkelley:
👍🏻 An example has been added.

hhsieh00 (approved):
Okay, all looks good now!  I can import State without error now.  I also tried implementing the ejection velocity example and was able to get it run, although the example synchrones for a 10km ejection velocity are all way out of frame of the sample plot (keeping everything else the same).  Using a 0.01 km sample ejection velocity keeps everything visible within the same dRA,dDec limits, but looks obviously different, so the added velocity clearly makes a difference.  So @mkelley, you might want to change the example ejection velocity in the documentation, but otherwise, seems good to go.

[test_synchrone_plot_with_delta_v.pdf](https://github.com/user-attachments/files/16667016/test_synchrone_plot_with_delta_v.pdf)

PR #395, opened 2024-01-26
Configure pep8speaks to use 88 character line length

mkelley:
I'm happy with black formatting, which uses an 88 character line length limit.  Astropy has also moved to formatting code with black.  Shall we change pep8speaks to use an 88 character limit?  (see #394 for an example with 80 character line length but code formatted with black).


hhsieh00 (approved):
sounds good

Issue #397, opened 2024-07-09
Tests: 2 errors during collection: astropy.units.core.UnitConversionError: '' (dimensionless) and 'Angstrom' (length) are not convertible


Hellseher:
**High-level problem description**
While packaging sbpy for Guix I've got issue to pass tests.

**What did you do?**
I prepared package for Guix.

**What did you expect?**
Tests passed.

**What did really happen?**
Tests failed during collection phase.
Trace:
```
sing pytest
============================= test session starts ==============================
platform linux -- Python 3.10.7, pytest-7.1.3, pluggy-1.0.0 -- /gnu/store/jh59fh48mcffyz5wmsjj0p96xkkflbz0-python-wrapper-3.10.7/bin/python
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/tmp/guix-build-python-sbpy-0.4.0.drv-0/sbpy-0.4.0/.hypothesis/examples')
rootdir: /tmp/guix-build-python-sbpy-0.4.0.drv-0/sbpy-0.4.0, configfile: tox.ini
plugins: hypothesis-6.54.5, astropy-0.11.0, doctestplus-1.2.0, remotedata-0.4.1, xdist-2.5.0, openfiles-0.6.0, filter-subpackage-0.2.0, cov-3.0.0, astropy-header-0.2.2, arraydiff-0.6.1, forked-1.6.0, asdf-3.2.0
collecting ... collected 626 items / 2 errors

==================================== ERRORS ====================================
__________ ERROR collecting sbpy/spectroscopy/tests/test_specgrad.py ___________
/gnu/store/5426ln4wv4pxriyp5bw8dg4wssrbr3ji-python-astropy-6.1.1/lib/python3.10/site-packages/astropy/units/quantity.py:987: in to_value
    scale = self.unit._to(unit)
/gnu/store/5426ln4wv4pxriyp5bw8dg4wssrbr3ji-python-astropy-6.1.1/lib/python3.10/site-packages/astropy/units/core.py:1191: in _to
    raise UnitConversionError(f"'{self!r}' is not a scaled version of '{other!r}'")
E   astropy.units.core.UnitConversionError: 'Unit(dimensionless)' is not a scaled version of 'Unit("Angstrom")'

During handling of the above exception, another exception occurred:
sbpy/spectroscopy/tests/test_specgrad.py:9: in <module>
    class TestSpectralGradient():
[...]

PR #400, opened 2024-07-10
Bugfix bandpass, address scipy.integrate.trapz deprecation, address astropy security vulnerability

mkelley:
* scipy 1.6 renamed `scipy.integrate.trapz` to `scipy.integrate.trapezoid`, and scipy 1.14 has dropped `scipy.integrate.trapz`.  Limit support to scipy>=1.6 (release Dec 2020) and use scipy.integrate.trapezoid.
* astropy<5.3.3 has a critical security vulnerability: require astropy >= 5.3.3
* Update min numpy version to match astropy 5.3
* synphot 1.4 deprecated the `wave_unit` keyword, which causes our tests to crash when reading in the SDSS filter transmission files: add TUNIT1="ANGSTROM" to the filter transmission FITS file headers and only use `wave_unit` when reading in the PanSTARRS filter transmission files.
* Testing with pyoorb and numpy=2 causes some tests to fail.  I think this is due to a difference in the ABI versions.  Limit continuous integration testing with pyoorb to numpy<2.

addresses #397


mkelley:
I think the failed code coverage check can be ignored.  I removed a lot of lines from the documentation that were for astropy < 5.3, so the total number of lines covered by the tests has gone down.

hhsieh00 (approved):
Looks good to me

PR #401, opened 2024-07-11
Use pip installable pyoorb

### LINKED ISSUE #384: Use pip installable pyoorb for testing?
pyoorb is now pip installable.  Using pip will help with testing against specific oorb versions.  The data is not packaged with the pypi package, so some thought is needed on how to get the data into the testing environment.

mkelley:
Also, add a new helper function (`sbpy.tests.helpers.get_pyoorb_data`) to install oorb data files needed for CI testing.

Closes #384


mkelley:
Needs to be rebased after #400 is merged.

hhsieh00 (changes requested):
I'm seeing several failed tests under checks for Python 3.10 and 3.11.  Among other things, looks like there are some exceptions in photometry/iau.py that need better handling and some missing or mis-linked data files for oorb.  codecov/patch is also not hitting its specified target, but not sure if this is a critical problem.

mkelley:
I'm actually stuck on this one.  I need to set the OORB_DATA environment variable, but I can't figure out how to get the correct path for the Oorb data, which I download into the virtual environment's share directory.  I'm going to try to find someone that can help.

mkelley:
OK, the oorb data now downloads and is made available to the testing environments using Github actions caching.  A job runs at the start of every CI test run that checks for the oorb ephemeris, time, and observatory code data files.  If they have not been previously cached, those files are downloaded and cached.  The cache shows up as "oorb-data" in the Actions section (see Management: Caches).  When the actual testing environments run, the cache is restored.

We run our tests using the OpenAstronomy github actions workflows.  However, the macos tests could not find the oorb data, even though it was restored from the cache.  I tried to understand this, but ultimately moved the macos tests to a separate job, and the tests work there.  🤷🏻 

The ephemeris data was previously built specifically for sbpy testing.  It is JPL's DE430 planetary ephemeris, but limited in time (1990 to 2030, see https://github.com/mkelley/pyoorb-experiment/tree/sbpy-testing ).  I think we can build the full ephemeris data set and cache that, but I'd like to save that work for the future. 

I've also re-written the weekly tests to use the OpenAstronomy workflow.  There is a link check, a test using dev versions of key dependences, and a remote data test.  I've marked the latter as "allowed_failures" given the transient nature of some internet resources.

hhsieh00 (approved):
Looks good now

mkelley:
Thanks!

PR #404, opened 2024-07-23
Actions: newly opened issues and pull requests

mkelley:
Based on Astropy's actions.  Links for more info still mostly direct people to Astropy, which is fine because we are an affiliated package that should follow their guidelines, too.


hhsieh00 (approved):
Looks good!

PR #413, opened 2024-09-11
Deprecate ginga_plugins.

mkelley:
Deprecate the `CometaryEnhancements` class and the `ginga_plugins` submodule itself.  Also, add a deprecation message to the `CometaryEnhancements` gui, in case users are not watching the console.  There is no documentation to edit, as this is a sort of pre-release feature that was never formally documented.

Instead use the sbpy-ginga module: https://github.com/NASA-Planetary-Science/sbpy-ginga

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,7 +14,7 @@ sbpy.names
 API Changes
 -----------
 - Deprecated `sbpy.ginga_plugins` in favor of using `sbpy-ginga` at
-  https://github.com/NASA-Planetary-Science/sbpy-ginga
+  https://github.com/NASA-Planetary-Science/sbpy-ginga [#413]
 
 
 0.5.0 (2024-08-28)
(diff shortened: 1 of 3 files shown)


hhsieh00:
5 lines in CometaryEnhancements.py are marked as not being covered by tests (with the CodeCov report on this page as well as the CodeCov messages on the code diffs page actually consistent this time).  Is this a problem or no?

hhsieh00 (approved):
Sorry, just getting to this now.  I installed ginga, opened it, opened a file, and don't even see CometaryEnhancements in the plugin menu to check for a deprecation message, but if that's the case, it seems that no one should encounter this anyways, so I'm going to call it good.

mkelley:
If you open ginga with `ginga --stderr` is there a line like the following?
```
2024-09-21 08:34:43,806 | I | PluginManager.py:70 (load_plugin) | Plugin 'CometaryEnhancements' loaded.
```

Issue #408, opened 2024-08-27
Cometary designations are not fully supported.


mkelley:
Cometary designations are not fully supported:

```
In [22]: Names.to_packed("73P")
---------------------------------------------------------------------------
TargetNameParseError                      Traceback (most recent call last)
Cell In[22], line 1
----> 1 Names.to_packed("73P")

File /disks/data0/astro/Projects/sbpy/sbpy/data/names.py:194, in Names.to_packed(s)
    190             raise TargetNameParseError(
    191                 ('{} cannot be turned into a '
    192                  'packed number or designation').format(s))
    193 else:
--> 194     raise TargetNameParseError(
    195         ('{} cannot be turned into a '
    196          'packed number or designation').format(s))

TargetNameParseError: 73P cannot be turned into a packed number or designation

In [23]: Names.to_packed("P/2004 A1")
---------------------------------------------------------------------------
TargetNameParseError                      Traceback (most recent call last)
Cell In[23], line 1
----> 1 Names.to_packed("P/2004 A1")

File /disks/data0/astro/Projects/sbpy/sbpy/data/names.py:194, in Names.to_packed(s)
    190             raise TargetNameParseError(
    191                 ('{} cannot be turned into a '
    192                  'packed number or designation').format(s))
    193 else:
--> 194     raise TargetNameParseError(
    195         ('{} cannot be turned into a '
    196          'packed number or designation').format(s))

TargetNameParseError: P/2004 A1 cannot be turned into a packed number or designation

In [24]: Names.to_packed("C/2013 US10")
---------------------------------------------------------------------------
TargetNameParseError                      Traceback (most recent call last)
Cell In[24], line 1
----> 1 Names.to_packed("C/2013 US10")

File /disks/data0/astro/Projects/sbpy/sbpy/data/names.py:194, in Names.to_packed(s)
    190             raise TargetNameParseError(
    191                 ('{} cannot be turned into a '
    192                  'packed number or designation').format(s))
    193 else:
--> 194     raise TargetNameParseError(
    195         ('{} cannot be turned into a '
    196          'packed number or designation').format(s))

TargetNameParseError: C/2013 US10 cannot be turned into a packed number or designation
```

These examples should use the 12-character format at:

https://www.minorplanetcenter.net/iau/info/PackedDes.html

Take care to account for two characters in a comet fragment designation, at least for numbered periodic comets.  The MPC's comet orbit table file packs 73P-BA into `"0073P     ba"`:

[...]

PR #417, opened 2024-11-09
identify invalid comet designation with fragment

hhsieh00:
Added code to check to see if comet designation without a fragment is a valid designation or not (e.g., not passing 2024 A-A when 2024 A is not a valid designation).  Basically just checks to see if the remaining designation without the fragment suffix has enough characters to be turned into a normal packed designation.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,7 +29,7 @@ sbpy.names
 ^^^^^^^^^^
 - Fixed `sbpy.Names.to_packed()' to raise an error in cases of invalid
   cometary designations with fragment specifiers (e.g., "2024 A-A" and
-  "2024 A-AA" now correctly raise TargetNameParseError exceptions)
+  "2024 A-AA" now correctly raise TargetNameParseError exceptions) [#417]
 
 
 0.5.0 (2024-08-28)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,12 @@ sbpy.activity.gas
 - Replaced calls to the deprecated function `scipy.integrate.romberg` with
   `scipy.integrate.quad`.  [#412]
 
+sbpy.names
+^^^^^^^^^^
+- Fixed `sbpy.Names.to_packed()' to raise an error in cases of invalid
+  cometary designations with fragment specifiers (e.g., "2024 A-A" and
+  "2024 A-AA" now correctly raise TargetNameParseError exceptions)
+
 
 0.5.0 (2024-08-28)
 ==================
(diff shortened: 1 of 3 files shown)


hhsieh00:
For what it's worth, I didn't use the suggested fix approach listed in the originating issue (removing the fragment designation, parsing the designation using normal rules without the fragment, and adding the fragment back) because it would mean replicating the fragment parsing code in a second location, where invalid non-fragment designations (e.g., 2024 A) fail simply because the string is too short, so it seemed easier to just keep the single instance of the fragment parsing code and just testing the length of the remaining string and raising a failure if it's too short.

mkelley:
Are two character fragments supported by the standard?  If not, then I'm OK with this approach.

hhsieh00:
If by "supported by the standard", you mean that something like 2024 A-AA will also raise a TargetNameParseError, the answer is yes, but for a different reason than the failure of 2024 A-A. 2024 A-A will fail because the remaining designation after stripping the fragment designation (2024 A) is not long enough.  2024 A-AA will raise a TargetNameParseError as well, but because it actually currently would not be recognized as a comet at all.  For comets, the code currently looks for (1) a year + single letter + number OR (2) a single-character fragment at the end of the designation (i.e., a dash + single letter, with no condition on what precedes the fragment designation), so 2024 A-AA will fail both (1) and (2) and pass through to the asteroid designation parsing block.  There, "A-" will be interpreted as the half-month part of the designation, and the remainder of the string "AA" will be interpreted as the asteroid number in that half-month and fail because it's not numeric.

This would be the right outcome for the wrong reason, and could be fixed by initially screening for comets by nesting (2) inside (1) (only testing (2) if (1) passes), rather than the current "or" statement.  I can change the fix to that approach if you prefer.

Otherwise, I think dealing with two character fragments in general runs into [Issue 408](https://github.com/NASA-Planetary-Science/sbpy/issues/408), where two character fragments require the longer packed comet designation format since the 7-character packed comet designation format (which is the only format option in the current code) only has space for a single-character fragment, i.e., 2024 A1-Z will pass, but 2024 A1-AA will fail).  I can work on that first to get this all fixed up before getting into the more involved astorbDB query development.

mkelley (changes requested):
I think you can go ahead with this, as it solves the main issue.  But I'm thinking there is room for improvement, e.g., we could use the is_comet and is_asteroid tests for code branching, instead of testing precise string indices.

Add something to the changelog and this can be merged.

mkelley (approved):

PR #419, opened 2024-11-11
Update versions

mkelley:
I'm generally trying to follow the guidance at https://scientific-python.org/specs/spec-0000/

* Require python >=3.9 (more permissive than above)
* Require numpy >=1.24
* Require scipy >=1.10
* Update testing versions in tox.ini
  * Test python 3.9 - 3.13
  * Numpy 1.24 to 2.1
  * Use python 3.11 for build_docs and linkcheck (for newer versions we will need an updated pyoorb release)
  * matplotlib >= 3.8 to match python 3.9
* Updated github workflows to use python 3.13 and 3.11.
 
Also reviewed python 3.9 change log: replaced deprecated typing.Tuple and typing.List with tuple[] and list[].


mkelley:
Ready for review!

.pep8speaks.yml
mkelley, line 1: To explain this change: Two files were merged and the other was deleted.  I also added E203 to the ignore list because it conflicts with https://peps.python.org/pep-0008/#whitespace-in-expressions-and-statements

hhsieh00 (approved):
Looks good to me

Issue #421, opened 2025-02-06
Incorrect Classification of Names.asteroid_or_comet


noedgemoonwind:
The ```Names.asteroid_or_comet()``` function incorrectly classifies ```asteroid``` names as ```comet``` when **Unpacked Provisional Designations** of the asteroid does not contain any space. 

For example,

```
>>> import sbpy
>>> sbpy.__version__
'0.5.0'
>>> from sbpy.data import Names
>>> Names.asteroid_or_comet('2015 XN77')
'asteroid'
>>> Names.asteroid_or_comet('2015XN77')
'comet'
```

```2015XN77``` is a invalid designation, the function should raise a **TargetNameParseError** to indicate that the input cannot be parsed. Or it might be better to classify **Unpacked Provisional Designations**  that does not contain any space like ```2015XN77``` as ```asteroid```, considering the program's robustness and possible usage scenarios. 

sbpy version: 0.5.0
astropy version: 6.1.0

Thank you very much for your attention to this issue.

mkelley:
Thank you for reporting this.  We will add it to our test suite.  I prefer TargetNameParseError, since the standard is to have a space after the year.  Any preference from @hhsieh00 ?

hhsieh00:
Yes, thanks for catching that!  I agree with @mkelley that TargetNameParseError makes more sense, and that it should be the user's responsibility to provide well-formatted input.  Playing around, I note that "2015XN77" returns "comet" as noted, but also "P2015XN77" returns "asteroid" (although "P/2015XN77" and "P/2015 XN77" both return "comet"), so I think it could get unwieldy to have sbpy be robust against all possible designations with non-standard formatting that users might attempt to classify.

hhsieh00:
@noedgemoonwind , this should hopefully be fixed now.  Please let us know if you encounter any other issues.  Thanks!

PR #422, opened 2025-02-12
fix incorrect asteroid-comet classification bug

hhsieh00:
Fixing a bug where parse_comet() incorrectly interprets input like "2015XN77" as a well-formed comet designation (where number=2015, type=X, and name=N77)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,11 +35,17 @@ sbpy.names
   cometary designations with fragment specifiers (e.g., "2024 A-A" and
   "2024 A-AA" now correctly raise TargetNameParseError exceptions) [#417]
 
-- Fixed `sbpy.Names.parse_comet(()' to raise an error in cases of a
+- Fixed `sbpy.Names.parse_comet()' to raise an error in cases of a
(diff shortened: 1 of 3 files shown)


hhsieh00:
So the problem in issue #421 seems to actually be due to the second part of the asteroid designation also being a potential comet type, so parse_comet will interpret "2015XN77", "2015PN77", and "2015CN77" all as comets (broken down as number=2015, type=X/P/C, and name=N77), while "2015TN77" or any other letters in that first spot besides PDCX will (correctly) throw an error.  The simplest fix seems to be to apply a validation test to the comet name to disallow comet names with numbers in the second position (but allow other non-alphabetic characters so names like d'Arrest don't also get screened out).  I considered adding a validation test earlier on in the parsing to look for years followed by asteroid designations without spaces, but I feel like that has a better chance of introducing unintended consequences (especially once LSST takes us into the four-digit comet number era), whereas simply validating the name seems to address this specific issue without impacting any other functionality.

mkelley (approved):

mkelley:
Thanks!

PR #425, opened 2025-09-05
Photolysis rates update

mkelley:
* Add new photolysis rates for H2O and CO2 from Huebner & Mukherjee 2015.
* Reformat `photo_timescale` and `photo_lengthscale` exception messages to show a nice table.
* Require `source` for  `photo_timescale` and `photo_lengthscale` (and subsequently `productionrate.beta_factor`).
* Light code reformatting.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,22 +10,34 @@ Update supported versions [#419]:
 New Features
 ------------
 
+sbpy.activity.gas
+^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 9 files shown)


mkelley:
The codecov test fails because I reformatted the productionrate.py file, but that file doesn't have good coverage in the tests.  I think we can proceed without this passing.

hhsieh00 (approved):
Changes look good

PR #427, opened 2025-10-17
Infrastructure update

mkelley:
* Remove sbpy.cfg, moving material to pyproject.toml
* Remove astropy testrunner, addresses #423 
* Version bump python -> 3.10
* Minor license updates
* Logo v2
* Readthedocs updates (os and python version)
* Disabled testing with pyoorb for now (need a new release of pyoorb that supports python 3.12+)
* Run compile_fieldnames as a tox command with build_docs
* Use macos-latest in github actions
* Test with numpy 2
* Update package version handling to match astropy
* Update doctests for numpy>=2
* Drop `Union` in type annotations (python 3.10)
* Better handling of Self type annotations.  These should get revisited as our python minimum evolves.
* Update coverage options to match astropy
* Minor docstring updates (**kwargs is a dict)
* Update dev notes (we are now working on v0.7)

PR #431, opened 2025-12-12
Add LSST filters

mkelley:
```python
>>> from sbpy.photometry import bandpass
>>> u = bandpass("LSST u")
>>> u.avgwave()
<Quantity 3723.54596825 Angstrom>
```


mkelley:
Looks like MacOS tests will continue to fail until #427 is done.

mkelley:
This one is rebased and ready for review.  It should be easier than the others, @hhsieh00 !

hhsieh00:
The code looks good.  I was thinking though that it would be useful for the main documentation page to point out that a list of available filters, the syntax for specifying them, and references are available at the bandpass() function page.  Also optionally, would it be possible to add an actual example of how to use the bandpass function (besides just how to load it into a parameter), like how to convert a mag from one filter to another?

mkelley:
Sounds good.  I've made that comment and added an example to calculate the LSST r-band magnitude of the Sun.  I'm noticing that the documentation function links are not working.  That's unrelated to this PR, so I'll open an issue and look into it.

hhsieh00:
That's great.  Thanks!  Another thought...would it be possible to also show another example of calculating the magnitude of an object with a given spectral slope (on top of a solar spectrum)?  I feel like that would be a use case of fairly broad interest (e.g., to figure out the equivalent colors in LSST filters of an object with an independently measured spectral slope, like in the 3I LSST paper), and I feel like it's probably trivial but not immediately obvious to me how to do it.  If you think this would clutter up the page too much, this would be okay to omit though (although I'd still like to know how to do it!).

mkelley:
Of course, but I'm going to call that out of scope for this PR, as that would be better done in the spectroscopy documentation.  I'll open an issue with an example.

hhsieh00 (approved):
Fair enough. Okay, looks good then!

mkelley:
Thanks!

PR #447, opened 2026-05-06
Fix: synphot keyword argument deprecation

mkelley:
* Only use flux_unit and wave_unit with synphot's read_*_spec when requested.
* Respect the cache parameter on SpectralSource.from_file


mkelley:
Ready for review.  That uncovered line is covered by the remote tests.  Maybe we should enable them on PRs again.  (although a few of them are currently failing)

hhsieh00 (approved):
Looks good.  Thanks for the fix!

